PR #11, opened 2026-04-23
feat: Add run-session text CLI to start a test session for the app

1u1m1a:
Adds new command `run-session text` which can be used to start a test session with an app.

`cxas run-session text projects/<project-id>/locations/<location>/apps/<app-id>`

### DIFF
--- a/src/cxas_scrapi/cli/main.py
+++ b/src/cxas_scrapi/cli/main.py
@@ -25,6 +25,7 @@ from typing import Dict, List
 
 import pandas as pd
 
+from cxas_scrapi import Sessions
 from cxas_scrapi.cli.app import (
     app_branch,
     app_create,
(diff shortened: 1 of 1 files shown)


src/cxas_scrapi/cli/main.py
```diff
@@ -1046,6 +1071,22 @@ def _add_project_location_args(
...
     parser_run.set_defaults(func=run_eval)
 
+    # Parser for 'run-session'
+    parser_run_session = subparsers.add_parser(
+        "run-session",
+        help="Start an interactive text session with the agent.",
+    )
+    parser_run_session.add_argument(
+        "modality",
+        choices=["text"],
```
kmaphoenix, line 1081: @1u1m1a are we planning to add `audio` as a future choice?

1u1m1a: We can add audio in the future. I just wanted to call out this is text only.

PR #13, opened 2026-04-24
fix: Fix bugs in init-github-action

1u1m1a:
Fixes `init-github-action` 
* Install scrapi via pip install cxas-scrapi
* Use dash-case for CLI invocations


kmaphoenix (approved):
LGTM

PR #15, opened 2026-04-24
feat: Add simulations to Golden conversion

vishalghor:
- Implemented `export_results_to_golden` method in `SimulationEvals` to enable converting simulation results into Golden evaluation YAML files.
-  Added platform-based trace fetching via `ConversationHistory` for high-fidelity conversion of tool calls and agent events. - Implemented a robust fallback parser that extracts turns from local `detailed_trace` logs if platform fetching is unavailable. - Refactored `simulation_evals.py` to resolve model name collisions (aliased `GoldenConversation` and `GoldenConversations`) and satisfied "Zero Warnings" linter requirements.
- Added a comprehensive unit test suite in `tests/cxas_scrapi/evals/test_sim_to_golden.py` covering both platform-based and fallback conversion paths.


src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -693,3 +703,171 @@ def _run_job(tc, run_idx):
...
+                                        if (
+                                            tc_obj.action == tool_name
+                                            and tc_obj.output is None
+                                        ):
+                                            tc_obj.output = response
+                                            break
+                if current_turn:
+                    turns.append(current_turn)
+
+            except Exception as e:
```
kmaphoenix: @vishalghor let's break out this entire except block into another method so that:
1. It's easier to debug
2. We can write unit tests just for this portion of the code

src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -693,3 +703,171 @@ def _run_job(tc, run_idx):
...
+            turns = []
+            try:
+                conv_obj = ch.get_conversation(session_id)
+                conv_dict = type(conv_obj).to_dict(conv_obj)
+
+                current_turn = None
+
+                for p_turn in conv_dict.get("turns", []):
+                    messages = p_turn.get("messages", [])
+                    for msg in messages:
```
kmaphoenix: @vishalghor Can we break out this entire turn / message parsing section into its own method so we can write better tests and debug it? 

There are a ton of nested if/else statements here as well.  
It would be good if we can compartmentalize some of the parsing here as well.  
e.g. `parse_tool_call`, `parse_tool_response`, `parse_user_text`, etc.)

I imagine that we have several classes throughout the library that end up parsing this information from Conversation History or Tracing already.  
We should try to consolidate these so it's reusable.

kmaphoenix (commented):
@vishalghor let's see if we can break down some of the larger methods so it's easier to debug / troubleshoot and write tests for them.

src/cxas_scrapi/evals/simulation_evals.py
vishalghor: Acknowledged

kmaphoenix (changes requested):

src/cxas_scrapi/evals/simulation_evals.py
vishalghor: added the susggestions. refactored the new addition and existing to be more modularized.

deniscalin (approved):
LGTM

tests/cxas_scrapi/evals/test_sim_to_golden.py
kmaphoenix, line 1: @vishalghor we shouldn't create a new file here. The tests should just go into `/evals/test_simulation_evals.py` that already exists.

kmaphoenix (approved):
@vishalghor let's fix the test file and then everything else LGTM

tests/cxas_scrapi/evals/test_sim_to_golden.py
vishalghor: makes sense. merged them

PR #29, opened 2026-04-28
feat: support events via audio

AlexanderKYu:
Add support for event handling when running audio.

Additional chore: remove utcnow due to deprecation

### DIFF
--- a/src/cxas_scrapi/evals/turn_evals.py
+++ b/src/cxas_scrapi/evals/turn_evals.py
@@ -683,11 +683,24 @@ class TurnEvals:
                     # Multi-turn sequence
                     full_conversation_trace = []
                     for step in case.turns:
+                        user_input = step.user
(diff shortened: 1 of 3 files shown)


src/cxas_scrapi/evals/turn_evals.py
```diff
@@ -683,11 +683,24 @@ def run_turn_tests(
                     # Multi-turn sequence
                     full_conversation_trace = []
                     for step in case.turns:
+                        user_input = step.user
+                        event_name = step.event
+
+                        if (
+                            user_input
```
ygupta89: We can do this via a regex

AlexanderKYu: Done

kmaphoenix (approved):
LGTM

PR #38, opened 2026-04-30
Add run_voice_evals_continuous to have voice evaluation using cxas_sim_eval skill

amanuppal17208:


kmaphoenix:
We have an implementation in core that handles BidiRunSession
https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/src/cxas_scrapi/core/sessions.py#L174

And Evals uses this implementation to do audio based evals. 
Will close this one as it is quite outdated now.  
Please feel free to raise another PR if we are missing functionality.

PR #39, opened 2026-04-30
fix: (docs) changed underscores in terminal command params to hyphens

vamsichivukula:
The terminal commands in the quickstart orientation - https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/docs/guides/agent-development/index.md#quick-orientation has underscore in the command param for app_name instead of app-name and project_id instead project-id causing the command to break.


kranthikiran01:
Please update the PR title to fix the semantic PR rule. add a prefix like `fix (docs) :`

PR #44, opened 2026-05-01
feat: improve colab migration logs

deniscalin:
### Changes
* **Colab Widget Log Streaming:** Implemented a custom `OutputWidgetHandler` to automatically stream Python `logging` outputs directly into the Jupyter/Colab UI cell (previously, logs were only routing to the background `.log` file).
* **Log Decluttering:** Removed noisy `print()` statements and `DEBUG TOOLSETS PAYLOAD` dumps to keep the dashboard output clean.
* **Decouple Gemini API vs CXAS Model:** Hardcoded the offline generative tasks (2A, 2B, 2C) to strictly use `gemini-3.1-pro-preview` for high-quality instruction and tool generation, completely decoupling it from the model deployed to the CXAS sub-agents (which is controlled by the user's dashboard selection).

### DIFF
--- a/src/cxas_scrapi/migration/service.py
+++ b/src/cxas_scrapi/migration/service.py
@@ -97,7 +97,7 @@ class MigrationService:
             project_id=project_id,
             location=gemini_location,
             credentials=credentials,
-            model_name=default_model,
+            model_name="gemini-3.1-pro-preview",
         )
 
         self.exporter = ConversationalAgentsAPI()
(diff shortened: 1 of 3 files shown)


src/cxas_scrapi/migration/service.py
```diff
@@ -97,7 +97,7 @@ def __init__(
             project_id=project_id,
             location=gemini_location,
             credentials=credentials,
-            model_name=default_model,
+            model_name="gemini-3.1-pro-preview",
```
vishalghor, line 100: Is there a reason why the model is hardcoded here? `default_model` at the top should be used

vishalghor (approved):
Added a comment. Rest is LGTM

src/cxas_scrapi/migration/service.py
deniscalin: There are 2 models at play here: one is for the migrated sub-agents (default_model), and the other one is for the tool itself (the hardcoded model).

I agree that we should create more descriptive variables, and remove the hardcoding, will send a follow-up change at some point.

PR #65, opened 2026-05-06
fix: Add missing google-cloud-storage dependency

danyal-s:
The setup failed because google-cloud-storage was missing when importing gcs_utils. This change adds it to install_requires.

Change-Id: I0cafb208dc411b608be4d53e074225956a6a6964


kmaphoenix (approved):
LGTM

PR #87, opened 2026-05-12
chore: update telemetry tracking patterns

kmaphoenix:
* Core Telemetry Enhancement
  - Added support for a `user_agent_extension` parameter in the Common base class.
  - This allows all API calls made by the SDK to be traced back to specific skills or scripts by appending them to the default user agent string.
* gRPC and WebSocket Integration
  - Updated core clients (e.g. Sessions, etc.) to use a custom gRPC transport that injects the extended user agent into the `grpc.primary_user_agent header`.
  - Updated BidiSessionHandler to pass the extended user agent in WebSocket headers for Bidi sessions.
* Skill Usage Tracking
  - Updated all skill scripts across .`agents/skills/cxas-agent-foundry/` and `.agents/skills/cxas-sim-eval/` to pass their specific identifiers (e.g. `skill/cxas-agent-foundry/app-thresholds`) when instantiating SCRAPI resources like Apps, Sessions, and Evaluations.
* Test Enforcement
  - Updated the mock client factories in `tests/conftest.py` to enforce that clients are initialized with a custom transport, ensuring that all new and existing code adheres to this telemetry pattern.

### DIFF
--- a/.agents/skills/cxas-agent-foundry/scripts/app-thresholds.py
+++ b/.agents/skills/cxas-agent-foundry/scripts/app-thresholds.py
@@ -36,11 +36,13 @@ from config import load_app_name, load_config, get_project_path
 HALLUCINATION_VALUES = {"unspecified": 0, "disabled": 1, "enabled_strict": 2, "enabled": 2}
 EXTRA_TOOL_VALUES = {"allow": 1, "deny": 2}
 
+USER_AGENT_EXTENSION = "skill/cxas-agent-foundry/app-thresholds"
+
 
 def get_app(project, location):
     """Get the Apps client."""
     from cxas_scrapi.core.apps import Apps
-    return Apps(project_id=project, location=location, user_agent_extension="skill/cxas-agent-foundry/app-thresholds")
+    return Apps(project_id=project, location=location, user_agent_extension=USER_AGENT_EXTENSION)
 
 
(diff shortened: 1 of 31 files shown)


.agents/skills/cxas-agent-foundry/scripts/app-thresholds.py
```diff
@@ -40,7 +40,7 @@
 def get_app(project, location):
     """Get the Apps client."""
     from cxas_scrapi.core.apps import Apps
-    return Apps(project_id=project, location=location)
+    return Apps(project_id=project, location=location, user_agent_extension="skill/cxas-agent-foundry/app-thresholds")
```
ygupta89: Should these be file level constants so we can change them easily in the future?

ygupta89 (changes requested):

.agents/skills/cxas-agent-foundry/scripts/app-thresholds.py
kmaphoenix: Yeah I like that idea.
let me run through and make those updates

kmaphoenix: @ygupta89 ok PTAL, this push is done

ygupta89 (approved):

PR #96, opened 2026-05-13
ci: declare workflow-level permissions on the two CI workflows

arpitjain099:
Two workflows declared no `permissions:` block.

- `ci.yml` — ruff + pytest, pure read. `contents: read` covers checkout.
- `commitlint.yml` — uses `amannn/action-semantic-pull-request@v5` which only reads the PR title to check conventional-commit format. `pull-requests: read` is the minimum needed. The `pull_request_target` trigger runs the action in the default-branch context, and the scope keeps it strictly read-only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ main ]
 
+permissions:
+  contents: read
+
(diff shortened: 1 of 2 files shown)


kmaphoenix (approved):
LGTM

PR #125, opened 2026-05-14
feat: Insights loss analysis skill

danielbeaudreau:
Adds a basic sim eval generator script which pulls successful conversations from CCAI insights API and then uses gemini to create evals from them.


danielbeaudreau:
(feedback: avoid gemini call, harness should make the evals itself)

danielbeaudreau:
fixes applied, @ygupta89 PTAL when you get a chance

.agents/skills/cxas-insights-sim-eval/scripts/generate_evals.py
```diff
@@ -0,0 +1,208 @@
...
+            self.creds, "token", None
+        ):
+            self.creds.refresh(Request())
+        return {
+            "Authorization": f"Bearer {self.creds.token}",
+            "Content-Type": "application/json",
+            "x-goog-user-project": self.project_id,
+        }
+
+    def get_conversations(
```
kmaphoenix: @danielbeaudreau any reason not to use the existing [ConversationHistory](https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/src/cxas_scrapi/core/conversation_history.py) class for this? Data should be identical.

.agents/skills/cxas-insights-sim-eval/scripts/generate_evals.py
```diff
@@ -0,0 +1,208 @@
...
+import requests
+
+# Ensure standard output is logging friendly
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
+)
+logger = logging.getLogger(__name__)
+
+
+class InsightsSimpleClient:
```
kmaphoenix: I don't think it's necessary to implement an Insights client for this task, however if we do end up needing it, we should extend the core [Insights](https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/src/cxas_scrapi/core/insights.py) class itself and then use the skill script to handle any extra wrapping logic.

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
```diff
@@ -0,0 +1,208 @@
...
+import google.auth
+from google.auth.transport.requests import Request
+import requests
+
+# Ensure standard output is logging friendly
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
+)
+logger = logging.getLogger(__name__)
+
```
kmaphoenix, line 33: If adding a new skill script, we need to ensure we're adding telemetry tracking.  See this script for detail / guidance:
https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/.agents/skills/cxas-agent-foundry/scripts/app-thresholds.py#L39

kmaphoenix (changes requested):

PR #129, opened 2026-05-15
fix: add widget tool linting and core SDK deserialization support

loso-code:
## Overview
This PR resolves widget tool handling limitations across both the **CLI Linter** and the **Core SDK** layer. It eliminates false positive linting errors on widget `.json` configurations and enables programmatic creation of widget tools via `Tools.create_tool`.

---

### 1. False Positive Linter Errors on Widget JSON Files
* **Symptom**: Running `cxas lint` on apps utilizing the new `widgetTool` capability reported compilation and returns-pattern errors (e.g., `T001`, `T010`) on widget `.json` tool manifests.
* **Root Cause**: Python-specific checks did not verify the file extension and attempted to compile/analyze JSON contents as Python source code.
* **Symptom**: `T012` falsely reported missing tool descriptions for widget tools.
* **Root Cause**: `T012` was hard-coded to only inspect `pythonFunction.description`, completely missing `widgetTool.description`.

### 2. SDK `create_tool` Deserialization Failure for Widgets
* **Symptom**: Creating a widget tool programmatically via the python client library (`Tools.create_tool`) failed because nested dictionary structures (like schemas inside `widgetTool.parameters`) were not correctly converted into Protobuf `google.protobuf.Struct` message fields.
* **Root Cause**: The core SDK relied on standard dictionary unpacks (`**kwargs`) which do not recursively handle complex Protobuf field conversions.

---

### Linter Layer (`src/cxas_scrapi/utils/lint_rules/tools.py`)
* **Suffix Guards**: Added explicit `.suffix != ".py"` guards to Python-only rules (`T001`, `T003`, `T005`, `T006`, `T009`, `T010`, `T011`) to cleanly skip JSON tool manifests.
* **Dynamic Description Resolving (`T012`)**: Refactored the check to evaluate whichever configuration block is active:
  * Inspects `pythonFunction.description` for standard Python tools.
  * Inspects `widgetTool.description` for widget tools.
* **Linter Unit Tests**: Added `test_t012_python_function_description`, `test_t012_widget_tool_description`, `test_t001_json_tool_skipped`, and `test_t010_json_tool_skipped` to prevent regressions.

### Core SDK Layer (`src/cxas_scrapi/core/tools.py`)
* **Protobuf Deserialization**: Upgraded `Tools.create_tool` to use `google.protobuf.json_format.ParseDict` to cleanly parse nested JSON configuration blocks into Protobuf messages.
* **Mock Backwards-Compatibility**: Added a fallback guard utilizing `unittest.mock.Mock` introspection. If a mocked `types.Tool` class is passed (during unit testing), it uses traditional attribute assignment, ensuring 100% backwards-compatibility with the existing mock-based test suite.

---

### 1. CLI Linter Verification
Ran `cxas lint` against a local widget-enabled agent project directory:
```bash
uv run cxas lint --app-dir /path/to/widget-agent

Before changes: Failed with 4 false-positive Errors (T001 and T012 errors on JSON tool manifests).
After changes: Passed cleanly with 0 errors.
2. Automated Test Suite
Ran the entire Pytest suite locally using uv run pytest:

Linter Tests: Passed successfully (114/114 passed).
Core Test Suite: Passed successfully (549/549 passed, 1 skipped).

### DIFF
--- a/src/cxas_scrapi/core/tools.py
+++ b/src/cxas_scrapi/core/tools.py
@@ -264,21 +264,16 @@ class Tools(Apps):
             )
             return self.client.create_toolset(request=request)
         else:
-            from unittest.mock import Mock  # noqa: PLC0415
-
             from google.protobuf import json_format  # noqa: PLC0415
 
             if description and "description" not in payload_copy:
                 payload_copy["description"] = description
 
             tool = types.Tool(display_name=display_name)
-            if isinstance(tool, Mock):
-                setattr(tool, tool_type, payload_copy)
-            else:
-                tool_dict = {tool_type: payload_copy}
-                json_format.ParseDict(
-                    tool_dict, tool, ignore_unknown_fields=True
-                )
+            tool_dict = {tool_type: payload_copy}
+            json_format.ParseDict(
+                tool_dict, tool._pb, ignore_unknown_fields=True
+            )
 
             request = types.CreateToolRequest(
                 parent=self.app_name, tool_id=tool_id, tool=tool
--- a/src/cxas_scrapi/core/tools.py
+++ b/src/cxas_scrapi/core/tools.py
@@ -265,6 +265,7 @@ class Tools(Apps):
             return self.client.create_toolset(request=request)
         else:
             from unittest.mock import Mock  # noqa: PLC0415
+
             from google.protobuf import json_format  # noqa: PLC0415
 
             if description and "description" not in payload_copy:
@@ -275,7 +276,9 @@ class Tools(Apps):
                 setattr(tool, tool_type, payload_copy)
             else:
                 tool_dict = {tool_type: payload_copy}
-                json_format.ParseDict(tool_dict, tool, ignore_unknown_fields=True)
+                json_format.ParseDict(
+                    tool_dict, tool, ignore_unknown_fields=True
+                )
 
             request = types.CreateToolRequest(
                 parent=self.app_name, tool_id=tool_id, tool=tool
--- a/src/cxas_scrapi/core/tools.py
+++ b/src/cxas_scrapi/core/tools.py
@@ -264,11 +264,19 @@ class Tools(Apps):
             )
             return self.client.create_toolset(request=request)
         else:
+            from unittest.mock import Mock  # noqa: PLC0415
+            from google.protobuf import json_format  # noqa: PLC0415
+
             if description and "description" not in payload_copy:
                 payload_copy["description"] = description
 
-            kwargs = {"display_name": display_name, tool_type: payload_copy}
-            tool = types.Tool(**kwargs)
+            tool = types.Tool(display_name=display_name)
+            if isinstance(tool, Mock):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/cxas_scrapi/core/tools.py [hunk 1]
```diff
@@ -264,11 +264,19 @@ def create_tool(
             )
             return self.client.create_toolset(request=request)
         else:
+            from unittest.mock import Mock  # noqa: PLC0415
```
kmaphoenix: @loso-code What's the purpose of using a mock here?

kmaphoenix (changes requested):
@kranthikiran01 can you PTAL for the linter rule changes

kmaphoenix:
@loso-code Nit: since this is not a net-new feature, please update the PR title to `fix:` instead so we can categorize the changes in the auto release properly

src/cxas_scrapi/core/tools.py
loso-code: @kmaphoenix  ParseDict was used to parse widget JSON schema. However, because the existing tests mock out the Tool class, ParseDict would crash during test execution.

Checking for Mock was the least invasive way to let the tests pass without doing a larger refactor of all the tool tests. The goal was to just add widget testing only.

src/cxas_scrapi/core/tools.py
```diff
@@ -264,11 +264,22 @@ def create_tool(
...
 
-            kwargs = {"display_name": display_name, tool_type: payload_copy}
-            tool = types.Tool(**kwargs)
+            tool = types.Tool(display_name=display_name)
+            if isinstance(tool, Mock):
+                setattr(tool, tool_type, payload_copy)
+            else:
+                tool_dict = {tool_type: payload_copy}
+                json_format.ParseDict(
+                    tool_dict, tool, ignore_unknown_fields=True
```
kmaphoenix: @loso-code you should be able to access `tool._pb` and avoid the need for the mock above.

```py
tool = types.Tool(display_name=display_name)
tool_dict = {tool_type: payload_copy}
json_format.ParseDict(
  tool_dict, tool._pb, ignore_unknown_fields=True
)
```

kmaphoenix: Then in `test_tools.py` we can make the following updates for `test_create_tool`:
- Drop the tool patch `@patch("cxas_scrapi.core.tools.types.Tool")`
- Change the test signature `test_create_tool(mock_client_cls, mock_req_cls)`
- Drop the final side effect mock `mock_tool_cls.side_effect = side_effect`

This should use the real `Tool` proto instead of the mock and avoid the failure.

kmaphoenix (changes requested):

src/cxas_scrapi/core/tools.py
loso-code: thank you much for the tip! Applied the refactor in those areas as you suggested.

kmaphoenix (approved):

PR #130, opened 2026-05-15
feat: split DFCX→CXAS skill into resumable migrate/stage1/2/3 scripts

vishalghor:
Adds the `cxas-dfcx-migration` skill: a four-script pipeline for migrating Dialogflow CX (DFCX) agents into CXAS (Customer Experience Agent Studio) apps, with optional Gemini-driven consolidation of fragmented agents into a smaller set of capability-rich groups. Also adds `cxas_scrapi.migration.structural_consolidator` so the same consolidation logic is callable from the CLI / notebooks / future skills.

Migrating a non-trivial DFCX agent to CXAS involves several distinct concerns — fetching + compiling the source, deduplicating variables, optionally collapsing many small agents into a few capability-rich ones, restructuring instructions into PIF XML, injecting tool mocks, wiring parent-child topology, generating tests, and linting the result. Doing all of that in a single command is slow (~30-60 min for a ~40-flow agent), brittle (any failure wastes the whole run), and opaque (no recovery point, no per-stage audit trail).

This skill splits the work into four small scripts that share state via a `<target>_ir.json` bundle on disk, so each stage can run / re-run / resume independently against an already-migrated agent without paying the per-flow LLM compile cost again.

- **NEW** [`structural_consolidator.py`](src/cxas_scrapi/migration/structural_consolidator.py) — Gemini-driven N→M agent consolidation, exposed as `StructuralConsolidator` plus pure helpers (`detect_root_key`, `validate_groupings`, `consolidate`, `rewrite_agent_refs`, `persist_grouping`, `load_grouping`). Synthesis is wrapped in `asyncio.wait_for(timeout=600s)` per group so a single hung Gemini call doesn't block the rest of the pipeline. Distinct from `CXASOptimizer` (which dedupes variables and restructures single-agent instructions); this module operates at the agent-set level.
- **FIX** [`optimizer.py:_stage1_variable_optimization`](src/cxas_scrapi/migration/optimizer.py) — escape literal `{` `}` braces around variable names in the substitution regex. Gemini legitimately emits numeric variable names (e.g. `'15'`); without escaping, `{15}` was parsed by Python's `re` as the `{n}` quantifier and crashed Stage 1 with `nothing to repeat at position 0`.


deniscalin (approved):
LGTM

PR #142, opened 2026-05-18
feat: implement strict local GECX linter rules and scaffolder guidelines

ygupta89:


src/cxas_scrapi/utils/lint_rules/config.py
```diff
@@ -215,3 +215,92 @@ def check(
...
+
+        results = []
+
+        # 1. Check case sensitivity / incorrect snake_case
+        if "root_agent" in data:
+            results.append(
+                self.make_result(
+                    file=rel,
+                    message=(
+                        "Found 'root_agent' in app.json, but GECX strictly "
```
kmaphoenix: nit: replace any `GECX` -> `CXAS` since this library only affects CX Agent Studio

kmaphoenix (approved):

src/cxas_scrapi/utils/lint_rules/config.py
ygupta89: done, and in other places in the file.

PR #152, opened 2026-05-19
feat: Consolidate and improve the usability protocols, and refactor the scoping layer to be cleanly structured as a first-class agent.

danyal-s:
- Added sub-skills not added in the first iteration into the CUJ Report Generator:
  - robust requirements extraction -> this skill is for comprehensive extraction of files from input files.
  - two-phase ingestion mapping -> this skill is used to unify file types into one digestible one to optimize reasoning during the transcript generation phase.
  - deterministic task checklist tracking -> this skill is used to optimize and prevent agent execution drifts.
    - this skill contains evals that can be run in an executor-evaluator pattern by an orchestrator. The eval data included tests various scenarios to give confidence in the skill's ability to handle multiple scenarios:
      -  3 files
      - 173 files
      - 1032 files
      - 10423 files
    These numbers were intentionally chosen this way to prevent the LLM from 'assuming' a number and to granuarly test how an LLM does with this skill when there are 3, 173, 1032, and 10423 checklist items. One may use the executor-evaluator pattern here to see how this skill fares for their LLM agent of choice.
- Integrate the scoper flow directly into the main skill's scoping step to establish a clean and declarative orchestration.


.agents/skills/cxas-cuj-report-generator/protocols/task-coverage-protocol/evals/cases/add_items_1032/case.yaml
```diff
@@ -0,0 +1,7 @@
+id: add_items_1032
```
kmaphoenix, line 1: @danyal-s is there a reason all of these case files are being added? if these are files generated as part of running a skill, we should exclude generated files and only commit the skill itself.

kmaphoenix (changes requested):

.agents/skills/cxas-cuj-report-generator/protocols/task-coverage-protocol/evals/cases/add_items_1032/case.yaml
danyal-s: These are intended eval cases used to benchmark this skill.

kmaphoenix (approved):
Discussed at length with @danyal-s offline.  
This will be the scaffolding for a new approach to requirements gathering. 
We'll merge functionality and docs with `cxas-agent-foundry` in upcoming PRs.

Issue #136, opened 2026-05-16
bug: Voice channel simulations do not evaluate against the returned audio.


sid-dinesh94:
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**
Voice channel simulations 

The final expectation check (such as checking if a tool was called, or if specific info was conveyed) is not performed directly on the audio output. Instead, the text-based Gemini LLM judge runs text-matching or logical evaluations on the reconstructed text transcript logs.

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:

Run cxas-scrapi's simulation eval script with --channel=audio against an agent that is instructed to incorrectly transcribe the audio

**Expected Behavior:**
The simulation evaluator should identify mismatched audio. One  function of  CXAS agent is to operate in voice mode. When we simulate in the voice channel, we are assuming that the transcript is correct, but often it is not.

**Observed Behavior:**
The expectation evaluator evaluates the ASR transcript against the expectation. We should use a multimodal model to evaluate against the output stream.

 **- Always (100%)**
 - Often (50%+)
 - Intermittently (<50%)
 - Once / Rare

P.S. I will have a Bug fix  / new feature flag to fix this soon.

sid-dinesh94:
The above PR is a part of the fix.

We will also need to inform through the skill references how and when to use this new feature.

PR #159, opened 2026-05-20
feat: Expectations can be judged based on audio token (`inline_data`)

sid-dinesh94:
Addresses https://github.com/GoogleCloudPlatform/cxas-scrapi/issues/136

**Introduces evaluate_expectations_with_audio_tokens which does**
 Allows the LLM-based User Simulator to perform evaluation of simulation test-cases directly on the raw, synthesized agent audio output rather than just the transcribed text. By passing raw audio files directly to the multimodal Gemini models, the simulator can verify auditory aspects of the GECX voice agent (such as pronunciation, intonation, and sound-quality expectations) under voice-to-voice/audio modality.

---

### How This is Done

**This is done by:**
1. **Audio Output Capture (`src/cxas_scrapi/core/sessions.py`)**:
   * Enhanced `BidiSessionHandler` to dynamically capture and stream raw audio bytes received during bidirectional voice sessions. 
   * Saves agent voice outputs to disk under `/tmp/scrapi_evals/{session_id}/turn_{turn_num}_agent.wav` when `evaluate_expectations_with_audio_tokens` is enabled, and returns the file paths in the `Sessions.run` response.
2. **Path Tracking & Accumulation (`src/cxas_scrapi/evals/simulation_evals.py`)**:
   * Updated `SimulationEvals` to track and accumulate agent audio file paths across conversational turns.
   * Propagates the active list of `agent_audio_paths` directly to the `LLMUserConversation` simulator context on every step of the conversation.
3. **Multimodal Token Assembly (`src/cxas_scrapi/utils/gemini.py` & `src/cxas_scrapi/prompts/llm_user_prompts.py`)**:
   * Configured the Gemini content generator to leverage its multimodal capabilities by reading local audio files and translating them into raw Gemini input audio tokens (via `_build_contents`).
   * Updated the system instructions for the LLM-based User Simulator to instruct the model on how to parse and evaluate the incoming audio tokens against expectations.
 4. **Simulation YAML Integration**
    * We add a new field `audio_expectations` to capture any expectations that should be evaluated against the audio tokens.

---

### Test Coverage

**Tests are added for:**
1. **Session Audio Handling (`tests/cxas_scrapi/core/test_sessions.py`)**:
   * Added `test_bidi_session_handler_audio_writing_enabled` to verify that the session handler correctly captures audio streams, generates WAV files, and populates `turn_audio_paths`.
   * Added `test_bidi_session_handler_audio_writing_disabled` to verify that no audio extraction or disk-writes occur when the feature is disabled.
2. **Simulation Integration (`tests/cxas_scrapi/evals/test_simulation_evals.py`)**:
   * Added `test_user_simulator_audio_with_eval_enabled` to verify that the simulator correctly orchestrates sessions with the `evaluate_expectations_with_audio_tokens=True` configuration and transfers accumulated audio paths successfully to the evaluator context.


.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -143,77 +141,128 @@ def simulate_conversation(
...
-                    }
-                    # Inject variables on first turn only
-                    if first_turn and session_params:
-                        kwargs["variables"] = session_params
-                        first_turn = False
-                    else:
-                        first_turn = False
-
-                    response = self.sessions_client.run(**kwargs)
+        try:
```
ygupta89: It's a bit weird to do this only in the skills instead of the base runner.

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -849,7 +935,252 @@ def cmd_run(args):
...
+    )
+
+    # Sensei Post-Eval
+    print("\nEvaluating conversation with Sensei...")
+    with open(args.doc, "r") as f:
+        instructions = f.read()
+
+    transcript_str = "\n".join(conv)
+
+    sensei_prompt = f"""You are an advanced AI quality auditor (Sensei) evaluating a conversation between a user (Dojo Agent) and a restaurant host agent.
```
ygupta89: We should put prompts in their own package for consistentcy

ygupta89 (changes requested):

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
sid-dinesh94: Apologies, I had some other changes for a prototype user model exercise bundled into this commit.

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
sid-dinesh94: Same here. I've reverted changes related to the experimental work.

kmaphoenix:
@ygupta89 can you PTAL and see if the latest changes are sufficient?

@sid-dinesh94 this needs a rebase with some of the latest file changes.

sid-dinesh94:
I updated the PR to do the following:

Audio expectations are now listed in an `audio_expectations` list at the same level as the `expectations` list. These are internally mapped to `expectations` with a `requires_audio` field set to true.

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -31,31 +31,26 @@
...
 from cxas_scrapi.utils.reporting import generate_html_report
 
-
-from config import load_app_name, get_project_path
-
 USER_AGENT_EXTENSION = "skill/cxas-agent-foundry/scrapi-sim-runner"
 
 
 EVALS_YAML = get_project_path("evals", "scenarios", "scenarios.yaml")
 SIM_EVALS_YAML = get_project_path("evals", "simulations", "simulations.yaml")
```
ygupta89, line 52: We can get rid of this path all together and have the DIR path only, similar to goldens.

src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -198,6 +208,18 @@ def __init__(
                 )
             )
         self.expectations = test_case.get("expectations", [])
+        audio_exps = test_case.get("audio_expectations", [])
+        self.audio_expectations = []
+        for exp in audio_exps:
+            if isinstance(exp, dict):
+                exp_dict = dict(exp)
+                exp_dict["requires_audio_paths"] = True
```
ygupta89, line 221: Whats this flag do? Why would it be false?

ygupta89 (changes requested):

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
sid-dinesh94: It can be a separate chore PR in the spirit of contained changes.

src/cxas_scrapi/evals/simulation_evals.py
sid-dinesh94: Added a comment here to clarify the use of this field internally. It is not exposed to users.

The field is set to false for text evals. Since we do not use Typed expectation structs, we do not have a way to differentiate between an evalset that needs the audio, and an evalset that does not (once they are parsed out of the YAML into SimulationEval fields). 

By setting it to false for text evals, we skip saving the audio tokens (memory and time) for text-token only evals.

ygupta89 (approved):

PR #162, opened 2026-05-20
feat: add cxas-loss-analysis skill for conversation loss and patterns analysis

danielbeaudreau:
Analyzes insights conversations and generates a report of loss patterns


.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
```diff
@@ -0,0 +1,232 @@
...
+
+    filter_arg = f'agent_id="{args.app_id}"'
+    logger.info(
+        f"Fetching recent conversations (target limit raw: {args.limit})..."
+    )
+
+    max_pages = (args.limit + 99) // 100
+    conversations = insights_client.list_conversations(
+        filter_str=filter_arg, page_size=100, max_pages=max_pages
+    )
```
sannat17: Why don't we allow the agent / user to define a time period at runtime and sample randomly from that?

Depending on the industry and agent ability, you may expect different traffic during different time of day, weekday vs weekends, etc.

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
```diff
@@ -0,0 +1,232 @@
...
+
+    conversations = conversations[: args.limit]
+    logger.info(f"Retrieved {len(conversations)} raw conversation summaries.")
+
+    # Filter for non-contained conversations (losses)
+    losses = []
+    for c in conversations:
+        contained = c.get("labels", {}).get("sessionContained")
+        if contained != "true" and contained is not True:
+            losses.append(c)
```
sannat17: Why don't we filter by this label within list_conversations? From what I see, the entire conversations sample is being discarded anyway after this subsampling.

In fact, we can allow arbitrary filter for the agent / user runtime to decide which subset of conversations (even within "losses") that they want to hill-climb / iterate on, and the current filter can be the default / suggested setup in the skill.

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
```diff
@@ -0,0 +1,232 @@
...
+    parser.add_argument(
+        "--limit",
+        type=int,
+        default=1000,
+        help="Max raw conversations to inspect (default: 1000)",
+    )
+    parser.add_argument(
+        "--loss-limit",
+        type=int,
+        default=100,
```
sannat17: Should have a higher default here, maybe 500 or 1000, since loss patterns follow a tail-end distribution in a lot of cases and may not have enough representation in a small sample.

sannat17 (changes requested):

src/cxas_scrapi/core/insights.py
```diff
@@ -113,3 +113,35 @@ def _list_paginated(
             if not page_token:
                 break
         return results
+
+    def list_conversations(
+        self,
+        filter_str: Optional[str] = None,
+        page_size: int = 100,
+        max_pages: int = 5,
+    ) -> List[Dict[str, Any]]:
```
sannat17, line 123: We can add the option to request the full conversation view instead of the basic default (which contains everything but the transcript) as is supported by the endpoint: https://docs.cloud.google.com/contact-center/insights/docs/reference/rest/v1/projects.locations.conversations/list

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
danielbeaudreau: Done

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
danielbeaudreau: Done

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
danielbeaudreau: Done

src/cxas_scrapi/core/insights.py
danielbeaudreau: Done

danielbeaudreau:
Thanks for the review! All comments should be addressed, so it is ready for a second pass.

.agents/skills/cxas-loss-analysis/scripts/fetch_losses.py
sannat17: We still have 2 different params `--limit` and `--loss-limit` even though we are sampling from losses directly in `list_conversations` now.  We can consolidate just to keep things simpler, unless you prefer the second subsampling to enable flexible slicing of the fetched data based on other heuristics / filters to identify specific loss patterns in batches.

danielbeaudreau: Done. That flag was previously left in for the sake of random sampling, but agree that it's probably adding more unnecessary complexity than value.

sannat17 (approved):

danielbeaudreau:
@ygupta89 can you please help merge this? thanks

ygupta89 (approved):

PR #175, opened 2026-05-22
fix: InquirerPy TUI event loop crash in grouping review

danyal-s:
Problem:
The Stage 1 migration script (`stage1.py`) crashes with `RuntimeError: asyncio.run() cannot be called from a running event loop` when attempting to launch the interactive grouping review TUI.

Root Cause:
The script is running inside an active `asyncio` event loop (`asyncio.run(_run(args))`). When it invokes `grouping_review.interactive_review` (which is called asynchronously), the TUI attempts to render InquirerPy prompts using the synchronous `.execute()` method. Under the hood, `.execute()` calls `asyncio.run()$, which strictly prohibits nesting within an already running event loop.

Solution:
Patched `grouping_review.py` and its prompt helper functions (`_merge_groups`, `_split_group`, `_rename_group`) to be asynchronous (`async def`) and use the native `await ...execute_async()` method instead. This allows InquirerPy to yield execution back to the existing active event loop instead of trying to spawn a nested one.

### DIFF
--- a/src/cxas_scrapi/migration/grouping_review.py
+++ b/src/cxas_scrapi/migration/grouping_review.py
@@ -145,7 +145,7 @@ def render_diff(
 # ---------------------------------------------------------------------------
 
 
-def _merge_groups(groupings: dict, console: Console) -> dict:
+async def _merge_groups(groupings: dict, console: Console) -> dict:
     names = list(groupings.keys())
     choices = [
         Choice(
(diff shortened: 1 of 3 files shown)


danyal-s:
This will take a bit more work because the unit tests are using MagicMock which don't work well with async. will standardize to not use MagicMock and instead prefer `create_autospec()`

vishalghor (approved):
LGTM

PR #176, opened 2026-05-22
feat: Add incremental HTML reporting during live simulation runs

hwhitney456:
Adds support for generating HTML evaluation reports iteratively as simulations progress, rather than waiting for the entire batch to finish. Includes a time-based throttle (15s) to avoid excessive disk writes during high-parallelism runs.


hwhitney456:
cc: @kmaphoenix - not sure how to add a reviewer. This is ready once someone has a moment (not urgent).

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -128,7 +129,11 @@ def simulate_conversation(
             test_case=test_case,
         )
 
-        session_params = test_case.get("session_parameters", {})
+        session_params = dict(test_case.get("session_parameters", {}))
+        # Inject a randomized virtual caller ANI to completely isolate session state storage across concurrent threads
+        session_params["ANI"] = f"888{random.randint(1000000, 9999999)}"
```
ygupta89, line 134: Why is this hard coded?

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -187,9 +193,13 @@ def simulate_conversation(
                         "escalat" in criteria
```
ygupta89, line 193: I actually think this is the wrong way of fixing this. We should fix the LLM user prompt instead.

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -377,7 +387,7 @@ def generate_html_report(
     if app_name:
         try:
             from cxas_scrapi.core.tools import Tools
-            tools_map = Tools(app_name=app_name, user_agent_extension=USER_AGENT_EXTENSION).get_tools_map()
```
ygupta89, line 380: Why delete this (and below)

src/cxas_scrapi/evals/runner.py
```diff
@@ -220,17 +221,22 @@ def run_all_evals(
...
-                        f"Running {len(test_cases)} simulations across "
-                        f"{len(sim_files)} files"
-                    )
-                    sim_results = sim_evals.run_simulations(
-                        test_cases,
-                        runs=runs,
-                        parallel=parallel,
-                        modality=modality,
-                    )
+                    if test_cases:
```
ygupta89, line 224: Why indent this?

ygupta89 (changes requested):

danyal-s:
Overall, I think it's important to separate progress reports from simulation reports. The former's purpose is for the developer/user to see progress and the latter is for sharable + explorartion of results.
Is it not enough to log progress reports in the console? Trying to understand the need for this.

danyal-s (changes requested):
left a comment for this.

kmaphoenix:
@hwhitney456 is this PR still valid?

Would be good if you can also address the comment here if we want to continue with this incremental style output.  
https://github.com/GoogleCloudPlatform/cxas-scrapi/pull/176#issuecomment-4606305233

Otherwise we can close the PR and consider what we have at HEAD right now.  
We've made a lot of changes since this was submitted.

PR #186, opened 2026-05-25
feat(cli): add cli commands for list versions and compare versions

royl627:
Introduce list version and compare version cli for scrapi

This feature gives user the ability to list all versions for an app. They could use the version ids fetched to compare the version to see human readable-diffs either in terminal console or html page generated

### DIFF
--- a/src/cxas_scrapi/cli/main.py
+++ b/src/cxas_scrapi/cli/main.py
@@ -40,6 +40,10 @@ from cxas_scrapi.cli.app import (
     apps_get,
     apps_list,
 )
+from cxas_scrapi.cli.versions_cli import (
+    app_versions_compare,
+    app_versions_list,
+)
 from cxas_scrapi.cli.create_local import handle_local_create
 from cxas_scrapi.cli.insights_cli import populate_insights_parser
 from cxas_scrapi.cli.migration_cli import (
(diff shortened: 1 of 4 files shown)


src/cxas_scrapi/cli/versions_cli.py
```diff
@@ -0,0 +1,895 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _upload_to_codebin(title: str, content: str) -> str | None:
+  """Uploads HTML content to Codebin using internal gosso and returns url."""
+  gosso_path = "/google/bin/releases/gosso/gosso"
+  if not os.path.exists(gosso_path):
+    return None
+
+  target_url = "https://codebin.googleplex.com/api/prototypes"
```
ygupta89: This is not available externally.

src/cxas_scrapi/cli/versions_cli.py
```diff
@@ -0,0 +1,895 @@
...
+
+    console.print(table)
+    console.print()
+
+  except Exception as e:
+    console.print(f"[red]Failed to list app versions: {e}[/]")
+    sys.exit(1)
+
+
+def _print_console_summary(
```
ygupta89, line 208: Can we use templates for these so we don't mix formatting code with output code

src/cxas_scrapi/cli/versions_cli.py
```diff
@@ -0,0 +1,895 @@
...
+        lines_html.append(
+            '<span style="color:#0366d6;background:#e1f5fe;'
+            f'display:block;font-weight:bold;">{escaped}</span>'
+        )
+      else:
+        lines_html.append(
+            f'<span style="color:#444;display:block;">{escaped}</span>'
+        )
+
+    block_html = f"""
```
ygupta89: Similar. Please use a template

ygupta89 (changes requested):

src/cxas_scrapi/cli/versions_cli.py
royl627: Do you think it is a good idea to create a local file and open as link in the browser?

royl627: If codebin failed or not available, the logic will fallback to generated local file instead

src/cxas_scrapi/cli/versions_cli.py
royl627: done

src/cxas_scrapi/cli/versions_cli.py
royl627: done

src/cxas_scrapi/cli/versions_cli.py
```diff
@@ -0,0 +1,929 @@
...
+{% if removed_toolsets %}  [red]➖ Removed :[/] {{ removed_toolsets }}
+{% endif -%}
+{% if modified_toolsets %}  [yellow]🔄 Modified:[/] {{ modified_toolsets }}
+{% endif -%}
+{%- endif %}
+[dim]* Use --verbose to print detailed line-by-line diffs to terminal[/]
+[dim]* Use --web to force open/view the full interactive HTML diff report[/]
+"""
+
+HTML_DIFF_BLOCK_TEMPLATE = """
```
ygupta89: i think we should this in another file. E.g.: https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/src/cxas_scrapi/utils/combined_report_template.html

ygupta89 (changes requested):

src/cxas_scrapi/cli/versions_cli.py
royl627: Done

src/cxas_scrapi/cli/main.py
```diff
@@ -1857,6 +1861,68 @@ def _add_project_location_args(
     )
     parser_local_create_tool.set_defaults(func=handle_local_create)
 
+    # Subparsers for 'versions'
+    parser_versions = subparsers.add_parser(
+        "versions", help="Manage GECX app versions (list, compare)."
```
ygupta89: Lets call this CXAS here and below.

ygupta89 (changes requested):

ygupta89 (approved):

PR #190, opened 2026-05-26
fix(trace): correct GECX/CES Console URL query parameter route

### LINKED ISSUE #189: Traces.console_url() builds wrong CES Console URL — no conversation loads
## Summary

`Traces.console_url()` builds a URL that doesn't map to a real CES Console route, so links produced by `cxas trace list` (the `ces_url` column), `cxas trace open`, and the report renderers all land on a 404-equivalent page instead of the conversation.

## Current behavior

`core/traces.py::console_url()` (around line 792):

```python
def console_url(self, conversation_id: str) -> str:
    base = self.trace_config.ui.ces_console_base.rstrip("/")
    return (
        f"{base}/projects/{self.project_id}/locations/{self.location}"
        f"/apps/{(self.app_name or '').split('/')[-1]}/conversations/"
        f"{conversation_id}"
    )
```

This produces:

```
https://ces.cloud.google.com/projects/<proj>/locations/<loc>/apps/<appId>/conversations/<convId>
```

…which is not a route the CES Console serves.

## Expected behavior

The CES Console renders a specific conversation via:

```
https://ces.cloud.google.com/projects/<proj>/locations/<loc>/apps/<appId>?panel=conversation_list&id=<convId>&source=<source>
```

`source` is the conversation's source enum (`LIVE` / `SIMULATOR` / `EVAL`) — without it, the panel may default to the wrong tab.

## Proposed fix

```python
def console_url(self, conversation_id: str, source: str | None = None) -> str:
    base = self.trace_config.ui.ces_console_base.rstrip("/")
    app_id = (self.app_name or "").split("/")[-1]
    params = {
        "panel": "conversation_list",
        "id": conversation_id,
    }
    if source:
        params["source"] = source
    qs = urllib.parse.urlencode(params)
    return (
        f"{base}/projects/{self.project_id}/locations/{self.location}"
        f"/apps/{app_id}?{qs}"
    )
```

Two call-site updates needed:

1. `Traces.list()` already has the source on each conversation — pass it through so the per-row `ces_url` carries `&source=…`.
2. `Traces.get_report()` and `Traces.bundle()` call `console_url(conversation_id)` without source. Either:
   - Look up the source from the normalized trace before building the URL, or
   - Make `source` optional on `console_url()` and accept the slightly-less-precise URL when it's not known.

## Impact

- `cxas trace list` table/json/csv (`ces_url` column)
- `cxas trace open <id>` (printed URL + macOS `open` shell-out)
- Markdown/HTML report rendering — every "Open in CES Console" link is broken
- Downstream tools that consume scrapi's URL (we hit this in the VS Code Trace Browser extension; mirror-fixed locally on our side)

## Repro

```sh
cxas trace list --app-name projects/<P>/locations/us/apps/<A> --app-dir <dir>
# Take any ces_url from the table → paste into a browser → CES UI doesn't load the conversation
```

kranthikiran01:
This PR corrects the GECX/CES Console URL routing constructed by Traces.console_url() to follow the query-parameter based format (?panel=conversation_list&id=<convId>&source=<source>), resolving 404/route errors when viewing conversations. All internal and CLI call-sites are updated to retrieve and pass the conversation source parameter.

Fixes #189

### DIFF
--- a/src/cxas_scrapi/cli/trace_cli.py
+++ b/src/cxas_scrapi/cli/trace_cli.py
@@ -424,7 +424,9 @@ def trace_bug_report(args: argparse.Namespace) -> None:
 def trace_open(args: argparse.Namespace) -> None:
     try:
         traces = _build_traces(args)
-        url = traces.console_url(args.conversation_id)
+        normalized = traces.get_normalized(args.conversation_id)
+        source_str = normalized.get("source")
+        url = traces.console_url(args.conversation_id, source=source_str)
     except Exception as e:
         print(f"Open failed: {e}", file=sys.stderr)
         sys.exit(1)
--- a/src/cxas_scrapi/core/traces.py
+++ b/src/cxas_scrapi/core/traces.py
@@ -33,6 +33,7 @@ import json
 import logging
 import os
 import subprocess
+import urllib.parse
 import zipfile
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from statistics import median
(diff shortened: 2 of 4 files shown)


vishalghor (approved):
LGTM

PR #194, opened 2026-05-27
feat: add voice background noise and randomized burst artifact injection

pankajkgarg:
Add support for continuous background noise and randomized burst artifact injection inside real-time voice sessions and simulation evaluations.

### Key Features
1. **Dynamic SNR Overlay Mixer:** Mixes continuous background hum and randomized burst artifacts into the synthesized TTS speech stream.
2. **Dynamic Volume Pre-scaling:** Computes speech and noise decibel levels (dBFS) at runtime using `pydub` to achieve the exact specified target SNR relative to user speech.
3. **Continuous Comfort Noise Streaming:** Streams comfort noise chunks chunk-by-chunk over the WebSocket bidirectional session during silences (pauses, trailing padding, thinking time), ensuring 100% acoustic continuity and avoiding audio phase resets.
4. **Graceful Fallback:** Bypasses effects gracefully if `pydub` is missing or audio assets are corrupt, returning clean voice bytes.
5. **Python 3.13+ Compatibility:** Includes `audioop-lts` as a version-constrained dependency to support systems running Python 3.13+.

### Verification
- Created full unit tests in `tests/cxas_scrapi/core/test_audio_transformer.py` covering SNR mixing, burst noises, and fallback scenarios (All tests pass 100% green).
- Manually verified active bidi stream noise injection over GECX locations in audio modality simulations.

### DIFF
--- a/src/cxas_scrapi/core/audio_transformer.py
+++ b/src/cxas_scrapi/core/audio_transformer.py
@@ -79,7 +79,9 @@ class AudioTransformer:
 
             if burst_noise_files and AudioSegment is None:
                 logging.warning(
-                    "burst_noise_files were provided, but pydub is not installed or failed to import (e.g., under Python 3.13+). Audio burst noise injection will be skipped."
+                    "burst_noise_files were provided, but pydub is not "
+                    "installed or failed to import (e.g., under Python 3.13+). "
+                    "Audio burst noise injection will be skipped."
                 )
 
             # Apply burst noise if provided and pydub is available
@@ -93,11 +95,11 @@ class AudioTransformer:
                     # Apply burst noise
                     for burst_file in burst_noise_files:
                         burst_noise = AudioSegment.from_file(burst_file)
-                        # Calculate dynamic volume to match target burst_noise_snr
-                        if (
-                            burst_noise.dBFS != float("-inf")
-                            and speech_audio.dBFS != float("-inf")
-                        ):
+                        # Calculate dynamic volume to match target
+                        # burst_noise_snr
+                        if burst_noise.dBFS != float(
+                            "-inf"
+                        ) and speech_audio.dBFS != float("-inf"):
                             target_burst_dbfs = (
                                 speech_audio.dBFS - burst_noise_snr
                             )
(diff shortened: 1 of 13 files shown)


src/cxas_scrapi/core/sessions.py
```diff
@@ -130,11 +136,68 @@ def __init__(
         self.ws_app = None
         self.outputs = []
 
+        # Setup continuous background noise segment if provided
+        self.bg_noise_segment = None
+        self.bg_noise_cursor = 0
+        if background_noise_file and AudioSegment is None:
+            logging.warning(
+                "background_noise_file was provided, but pydub is not installed or failed to import (e.g., under Python 3.13+). Audio background noise streaming will be skipped."
```
ygupta89: Should we throw an explicit error here since this will get ignored

src/cxas_scrapi/core/audio_transformer.py
```diff
@@ -58,6 +72,67 @@ def text_to_speech_bytes(
...
+                            )
+                        else:
+                            speech_audio = speech_audio.overlay(
+                                burst_noise, position=0
+                            )
+
+                    # Export the mixed audio to raw PCM bytes
+                    mixed_io = io.BytesIO()
+                    speech_audio = (
+                        speech_audio.set_frame_rate(16000)
```
ygupta89: Lets move 16000 to a constant

src/cxas_scrapi/core/audio_transformer.py
```diff
@@ -58,6 +72,67 @@ def text_to_speech_bytes(
...
+                        .set_channels(1)
```
ygupta89: Similar

ygupta89 (approved):

pankajkgarg:
I have addressed all the review comments:

1. **Sample Rate & Audio Properties Constants**: Extracted the hardcoded sample rate (16000), channels (1), and sample width (2) to shared module-level constants in `audio_transformer.py` (`AUDIO_SAMPLE_RATE_HZ`, `AUDIO_CHANNELS`, `AUDIO_SAMPLE_WIDTH`), and applied them consistently across both `audio_transformer.py` and `sessions.py`.
2. **Explicit ImportError for Missing pydub**: When background noise injection is explicitly requested (`background_noise_file` is provided) but `pydub` is missing, we now raise an explicit `ImportError` with setup instructions rather than silently continuing. A dedicated unit test `test_bidi_session_handler_pydub_missing_raises_error` has also been added to verify this behavior.

All local unit tests and formatting/lint checks pass successfully with these updates.

PR #198, opened 2026-05-27
feat: Surface system errors in HTML report by raising Gemini exceptions

hwhitney456:


kmaphoenix:
@hwhitney456 please update the corresponding unit tests to expect raising the errors vs. returning None which is what they are currently doing

otherwise the change LGTM

src/cxas_scrapi/utils/gemini.py
```diff
@@ -125,7 +125,7 @@ def generate(
             return response.text
         except Exception as e:
             logger.error(f"Gemini generation failed: {e}")
```
danyal-s, line 127: we should us logger.exception here to surface exceptions.

src/cxas_scrapi/utils/gemini.py
```diff
@@ -191,7 +191,7 @@ def generate_with_parts(
             return response.text
         except Exception as e:
             logger.error(f"Gemini multimodal generation failed: {e}")
```
danyal-s, line 193: same as above

src/cxas_scrapi/utils/gemini.py
```diff
@@ -263,7 +263,7 @@ async def generate_async(
                     logger.error(
```
danyal-s, line 263: same as above

danyal-s (changes requested):

PR #199, opened 2026-05-27
feat: Allow user to specify region when running simulations

hwhitney456:


src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -343,16 +343,17 @@ def __init__(
         self.app_name = app_name
         project_id = app_name.split("/")[1]
         location = app_name.split("/")[3]
+
+        # Vertex AI requires a specific region (e.g. global), whereas CXAS
+        # Apps use 'us' or 'eu'. Allow override via kwargs.
+        vertex_location = kwargs.pop("vertex_location", "global")
```
danyal-s, line 349: Instead of popping this from kwargs, let's support it in the args (with a default value)/make dependent functions gracefully ignore it.

danyal-s (changes requested):

src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -343,16 +343,17 @@ def __init__(
         self.app_name = app_name
         project_id = app_name.split("/")[1]
```
kmaphoenix, line 344: Use the prebuilt methods in core to get `project_id` and `location` safely
https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/src/cxas_scrapi/core/common.py#L167

kmaphoenix (changes requested):

PR #207, opened 2026-05-28
feat(migration): emit tabbed HTML analysis report on every migration

vishalghor:
Every `cxas migrate` (initial run, each stage, resume-from-bundle) now writes a self-contained `<target>_migration_analysis.html` next to the IR bundle. The report has six tabs — Overview, CXAS Agents, Tools & Variables, Evals (stubbed), DFCX Source, Pipeline — with a drawer-based inspector, global search, hash deep-linking, and a live Mermaid topology diagram. The file opens directly in a browser, requires no network beyond the Mermaid CDN, and updates atomically at each of 13 phase boundaries so users can watch the migration progress.


deniscalin (approved):
LGTM. Loving the way it turned out!

Super nit: to stay consistent with PEP 8 and assure readability, I would recommend using an underscore to separate stage number from `stage`, e.g., `stage_1` instead of `stage1`. I'm noticing this in variables, comments, code, etc.

I'll send a quick PR to rename these.

I have also pushed a fix for the .gitignore conflict to this PR.

PR #222, opened 2026-05-30
fix: add pre-commit hook to block brand names

vishalghor:
This update to pre-commit hook flag if any brand names are part of the commit.
Updated to CONTRIBUTING.MD to amke isntall of the precommit hook necessary

### DIFF
--- a/scripts/check_brands.py
+++ b/scripts/check_brands.py
@@ -189,7 +189,9 @@ def _build_user_prompt(
     numbered_lines: list[tuple[int, str]],
 ) -> str:
     allowed_csv = ", ".join(ALLOWED_BRANDS)
-    body = "\n".join(f"{lineno}: {content}" for lineno, content in numbered_lines)
+    body = "\n".join(
+        f"{lineno}: {content}" for lineno, content in numbered_lines
+    )
     return f"""Scan the following {scoped_label} for third-party brand /
 company / product names.
 
--- a/scripts/check_brands.py
+++ b/scripts/check_brands.py
@@ -69,8 +69,6 @@ ALLOWED_BRANDS: tuple[str, ...] = (
     "GECX",
     "CXAS",
     "DFCX",
-    "Cymbal",  # Google's fictitious-brand placeholder; includes variants
-    # like "Cymbal Telco", "CymbalRetail", etc.
     # --- Google products / services this repo legitimately references ---
     "Gemini",
     "Vertex AI",
--- a/scripts/check_brands.py
+++ b/scripts/check_brands.py
@@ -7,15 +7,13 @@
 #
 #     https://www.apache.org/licenses/LICENSE-2.0
 
-"""Pre-commit guard against third-party brand mentions, powered by
-Gemini.
+"""Pre-commit guard against third-party brand mentions.
 
-Replaces the earlier regex+allowlist approach. The hook collects the
-*added* lines of the staged diff (for ``--code-files``) or the in-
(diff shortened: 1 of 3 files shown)


scripts/check_brands.py
```diff
@@ -0,0 +1,422 @@
...
+    explicit raise to honor the hook's block-on-failure contract."""
+    project_id = _resolve_project_id()
+    if not project_id:
+        raise RuntimeError(
+            "GCP project not resolvable. Set GOOGLE_CLOUD_PROJECT or "
+            "run `gcloud config set project <id>`."
+        )
+
+    try:
+        from cxas_scrapi.utils.gemini import GeminiGenerate  # noqa: PLC0415
```
vishalghor: will move this to top of the file

scripts/check_brands.py
```diff
@@ -0,0 +1,422 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+
+"""Pre-commit guard against third-party brand mentions, powered by
+Gemini.
+
+Replaces the earlier regex+allowlist approach. The hook collects the
```
ygupta89: This looks like gemini thoughts that we can remove.

ygupta89 (dismissed):

scripts/check_brands.py
vishalghor: Done

scripts/check_brands.py
```diff
@@ -0,0 +1,427 @@
...
+
+# Brands the project IS allowed to mention. Sent verbatim in the prompt
+# so Gemini knows what NOT to flag. Lower-cased before comparison.
+ALLOWED_BRANDS: tuple[str, ...] = (
+    # --- Parent org + this project + our placeholder customer name ---
+    "Google",
+    "GECX",
+    "CXAS",
+    "DFCX",
+    "Cymbal",  # Google's fictitious-brand placeholder; includes variants
```
kmaphoenix: @vishalghor we can drop the comment here re: Cymbal

scripts/check_brands.py
```diff
@@ -0,0 +1,427 @@
...
+
+_GEMINI_SYSTEM = (
+    "You are a brand-name auditor for a Google Cloud open-source "
+    "Python project. Your sole job: scan the provided text for any "
+    "third-party company, brand, or commercial product name that is "
+    "NOT in the allowlist. Return STRICT JSON only."
+)
+
+
+def _build_user_prompt(
```
kmaphoenix, line 194: @vishalghor if we have a defined set of "approved" names, can we achieve the same result without using an LLM call to get the details?

The problem with making this an LLM call is that we force the user to consume LLM tokens on their project just to check their work because this is a pre-commit hook.

kmaphoenix (changes requested):

scripts/check_brands.py
vishalghor: with out an LLM we will likely have to maintain a brand blocked list apart from "approved" names. And this blocked list will expose internal customer engagements. This does add token cost but it will be minimal per commit ~(0.001$). also for those who would want to skip the precommit hook due to tokens they can use the `BRAND_CHECK_SKIP=1` in env to skip.

scripts/check_brands.py
vishalghor: Done

kmaphoenix (approved):

PR #225, opened 2026-05-30
feat: move llm-lint from a skill to a package prompt and register as a cxas CLI subcommand

monicalenart:
and register as a cxas CLI subcommand


kranthikiran01 (approved):

src/cxas_scrapi/cli/main.py
```diff
@@ -1898,6 +1899,36 @@ def _add_project_location_args(
     )
     parser_lint.set_defaults(func=app_lint)
 
+    # Parser for 'llm-lint'
```
ygupta89, line 1902: Should we have this as part fo the base cxas lint cli command that we can turn on/off with a --enable_llm_linter

ygupta89 (commented):
Thank you!

src/cxas_scrapi/cli/main.py
monicalenart: Actually I prefer to keep them separate as they use different compute (CPU versus TPU).

PR #227, opened 2026-05-31
feat(migration): Grouping Review tab in analysis report (read-only)

vishalghor:
Adds a new "Grouping Review" tab to the migration analysis report. The tab renders only when MigrationAnalysisSnapshot.pending_grouping is non-null, so existing report behavior is unchanged in production runs (no current code path sets the field; PR 4 will wire that).

Changes:

1. analysis_reporter.py - MigrationAnalysisSnapshot gains a pending_grouping: dict | None = None field, serialized through to the report-data blob.

2. analysis_report_template.html - Grouping Review tab (default display:none), panel-grouping section with toolbar / cards / sticky validation footer, CSS for .gr-card / .gr-pill / .gr-status-pill / .btn, and JS for renderGroupingReview, renderGroupingCards, renderGroupingBanner, renderGroupingValidation, validateGrouping, wireGroupingCardHandlers, wireGroupingToolbar. Cards expose labeled sections (Why grouped, What these flows do, Source flows). HTML5 drag-and-drop, inline rename, root toggle, delete, add, collapse, reset all work client-side.

   Confirm/Abort/Re-propose buttons render disabled with a "Live server not running" tooltip; they activate only when window.__REVIEW_ENDPOINT__ is set (PR 4 wires the endpoint).

3. analysis_report_template.html - (DATA.evals || {})[name] null-guard in renderAgentGrid and the agent drawer. Fixes a pre-existing TypeError that affected live migrations where the eval suite has not run yet; the unguarded access halted the BOOT chain. Net improvement to existing flows.

4. test_analysis_reporter.py - 5 new cases asserting pending_grouping defaults, serialization, tab visibility behavior, and that the Confirm button is disabled when no server endpoint is set.

### DIFF
--- a/src/cxas_scrapi/migration/analysis_report_template.html
+++ b/src/cxas_scrapi/migration/analysis_report_template.html
@@ -313,6 +313,45 @@
     body { background: white; color: black; }
     main { max-width: 100%; padding: 8px; }
   }
+
+  /* Grouping Review */
+  .btn { background: var(--panel); border: 1px solid var(--border); color: var(--fg);
+    padding: 6px 12px; border-radius: 6px; font-size: 13px; cursor: pointer; }
+  .btn:hover:not(:disabled) { border-color: var(--accent); }
+  .btn:disabled { opacity: 0.5; cursor: not-allowed; }
+  .btn-primary { background: var(--accent); color: var(--bg); border-color: var(--accent); }
+  .gr-card { border: 1px solid var(--border); border-radius: 8px; padding: 12px;
+    background: var(--panel); min-height: 120px; transition: border-color 0.15s; }
+  .gr-card.drag-over { border-color: var(--accent); background: var(--panel-hover, var(--panel)); }
+  .gr-card.root { border-left: 4px solid var(--accent); }
+  .gr-card-title { font-weight: 600; outline: none; border-bottom: 1px dashed transparent; }
+  .gr-card-title:focus { border-bottom-color: var(--accent); }
+  .gr-card-title.invalid { color: #d9534f; border-bottom-color: #d9534f; }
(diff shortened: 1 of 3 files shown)


deniscalin (approved):
**Approved to merge** as is (since the grouping tab remains hidden in standard runs and doesn't affect existing main execution paths). 

We should track and potentially address the following items in a follow-up PR once the main gate is wired:

1. **Draggable Orphan Flow Issue (Usability Block)**: 
   If the LLM consolidation proposal fails to include a flow (or if a user deletes a flow mapping manually via the plan JSON), `validateGrouping` flags this as a blocking "orphan flow" error. However, because unassigned flows are not rendered as draggable pills anywhere on the screen, the user is locked out from confirming and cannot drag the orphan flows into a group.
   * *Proposed Fix*: Render an "Unassigned Flows" container panel at the top or bottom of the grid containing draggable pills for all calculated orphans.

2. **XSS Vulnerability via DFCX Content Injection (Security)**:
   In `analysis_reporter.py`, `data_json` is serialized via `json.dumps` and injected raw into the template via `{{ data_json | safe }}` inside a `<script>` tag. If a Dialogflow CX agent has page/intent names or descriptions containing literal `</script>` tags (**not sure if that is even allowed in DFCX, but let's check**), the browser parser will close the JSON block early and execute arbitrary script commands.
   * *Proposed Fix*: Escape forward slashes (e.g. replacing `</` with `<\/`) or apply proper HTML escaping to the injected JSON string before rendering.

3. **Jarring Card Snapping on Invalid Rename (UI Glitch)**:
   When renaming a card to an invalid or duplicate name, the label is highlighted red (`.invalid`) but not reverted. If the user edits another card, a full card re-render is triggered, causing the first card to abruptly snap back to its original name.
   * *Proposed Fix*: Revert the text back to the original group name immediately on `blur` if name validation fails.

PR #228, opened 2026-05-31
fix: canonical-XML consolidation prompt + lint-driven retry

vishalghor:
The consolidation prompt (STEP_3B_CONSOLIDATION_INSTRUCTIONS) emits non-canonical CamelCase XML (<Agent>/<Conversation_Schema>/<state>) that diverges from docs/design-guide and every example agent, and fails cxas-lint rules I001/I002. This fix rewrites the prompt to emit the canonical lowercase taskflow schema and validates each synthesized agent inline so non-canonical XML can never ship.

Prompt fix:
  Canonical schema (<role>/<persona>/<primary_goal>/<constraints>/
  <guidelines>/<taskflow>/<subtask>/<step>/<trigger>/<action>) is
  documented explicitly and the Bella_Notte_Host/instruction.txt
  example is embedded verbatim as a few-shot (loaded at import time,
  with an inline fallback for installed wheels).

Validate-then-retry loop in StructuralConsolidator.synthesize_instructions:
  Each Gemini-synthesized group instruction is checked with
  lint_instruction_text; on diagnostics, designer.run_step_2b_instructions
  is re-called with a feedback block listing the violations. A second
  failure raises XMLSchemaError so non-canonical XML never ships.
  Re-prompt budget is exactly one to bound latency and token cost.

Rather than maintaining a parallel canonical-schema checker, reuse the existing cxas-lint engine:

  - New rule I015 (banned-legacy-xml-tags) in src/cxas_scrapi/utils/lint_rules/instructions.py — flags every legacy CamelCase / state-machine tag (<Agent>, <Conversation_Schema>, <state>, <transitions>, <Persona>, etc.). This is the one canonical- schema check the lint engine was missing; existing I001/I002 already cover required tags and taskflow children.

  - New helper src/cxas_scrapi/migration/instruction_lint.py wraps the registered instruction rules so they can be run against in-memory synthesis output. Curated to schema/syntax rules that don't need cross-file context (I001, I002, I010, I011, I015) — agent / tool registry rules (I008, I009) stay silent without a real project root.

Tests:
  - test_prompts.py — snapshot assertions that the Bella Notte example is embedded verbatim and that the canonical tag vocabulary is documented in the prompt.
  - test_consolidator_synthesis_retry.py — fake-Gemini retry cases: canonical first-try, legacy-then-canonical (retry succeeds), legacy-twice (raises XMLSchemaError with I015 diagnostics).
  - test_instruction_lint.py — helper passes canonical text, fires I015 on legacy text, does not false-positive on I008/I009 without cross-file context.
  - test_lint_rules.py — direct tests of the new I015 rule.


deniscalin (changes requested):
### 1. Avoid Procedural `If` Statements in Actions (`_CANONICAL_EXAMPLE_FALLBACK`)
In the inline fallback example (and the corresponding Bella Notte example file), the `Intent_Routing` subtask uses nested conditional checks within a single `<action>` block:
```xml
      <action>
        1. If the user wants to book a table, make a reservation... call {@TOOL: set_active_flow} with flow="reservation" silently.
        2. If the user wants to place a takeout order... call {@TOOL: set_active_flow} with flow="takeout" silently.
        3. If the user has general questions... respond conversationally.
      </action>
```
Under Programmatic Instruction Following (PIF) best practices, procedural branching inside a single action block degrades LLM execution accuracy. Instead, we should split these into separate declarative steps, using the `<trigger>` tag to handle the conditions:
```xml
  <subtask name="Intent_Routing">
    <step name="Route_To_Reservation">
      <trigger>Customer wants to book a table, make a reservation, or mentions reservation-specific details.</trigger>
      <action>Call {@TOOL: set_active_flow} with flow="reservation" silently.</action>
    </step>
    <step name="Route_To_Takeout">
      <trigger>Customer wants to place a takeout order or names a menu item to order.</trigger>
      <action>Call {@TOOL: set_active_flow} with flow="takeout" silently.</action>
    </step>
    <step name="Answer_General_Questions">
      <trigger>Customer has general questions about hours or location.</trigger>
      <action>Respond conversationally.</action>
    </step>
  </subtask>
```
I would recommend updating the `_CANONICAL_EXAMPLE_FALLBACK` string to use this declarative structure.

**As a follow-up PR, I will also consider suggesting updating `examples/bella_notte/agents/Bella_Notte_Host/instruction.txt`**, especially if we can test and validate that procedural branching performs better.

### 2. Evidence Supporting `<subtask>` & Lowercase XML Tags
I see that we are changing out `<state>` for `<subtask>`, along with a few other tags. I know that majority of agents uses `<subtask>`, but do we have evidence of `<subtask>` performing better than `<state>`? I have seen `<state>` work well for a customer agent.

### 3. XML Gating is Too Aggressive (Do Not Abort Compilation)
Currently, if the retry attempt fails schema validation, the consolidator throws an `XMLSchemaError`, which propagates through `asyncio.gather(..., return_exceptions=False)` and aborts the entire migration process.
Having non-`cxas lint` approved tags does not crash the agent at runtime. The migration tool should treat XML schema validation as a **non-blocking lint warning** rather than a compilation-halting error.
* **To fix this**: If the re-prompt still returns invalid XML:
  1. Log a loud warning (e.g. `logger.warning("Synthesized XML for %s still fails schema validation after retry. Proceeding anyway.")`).
  2. Set the status for that group in the returned dictionary to `"warning"`.
  3. **Save the generated instructions anyway** and allow the compilation to proceed.
  4. This avoids crashing the overall `asyncio.gather` process and ensures a single styling failure doesn't discard other successfully compiled agents.

vishalghor:
> ### 1. Avoid Procedural `If` Statements in Actions (`_CANONICAL_EXAMPLE_FALLBACK`)
> In the inline fallback example (and the corresponding Bella Notte example file), the `Intent_Routing` subtask uses nested conditional checks within a single `<action>` block:
> 
> ```
>       <action>
>         1. If the user wants to book a table, make a reservation... call {@TOOL: set_active_flow} with flow="reservation" silently.
>         2. If the user wants to place a takeout order... call {@TOOL: set_active_flow} with flow="takeout" silently.
>         3. If the user has general questions... respond conversationally.
>       </action>
> ```
> 
> Under Programmatic Instruction Following (PIF) best practices, procedural branching inside a single action block degrades LLM execution accuracy. Instead, we should split these into separate declarative steps, using the `<trigger>` tag to handle the conditions:
> 
> ```
>   <subtask name="Intent_Routing">
>     <step name="Route_To_Reservation">
>       <trigger>Customer wants to book a table, make a reservation, or mentions reservation-specific details.</trigger>
>       <action>Call {@TOOL: set_active_flow} with flow="reservation" silently.</action>
>     </step>
>     <step name="Route_To_Takeout">
>       <trigger>Customer wants to place a takeout order or names a menu item to order.</trigger>
>       <action>Call {@TOOL: set_active_flow} with flow="takeout" silently.</action>
>     </step>
>     <step name="Answer_General_Questions">
>       <trigger>Customer has general questions about hours or location.</trigger>
>       <action>Respond conversationally.</action>
>     </step>
>   </subtask>
> ```
> 
> I would recommend updating the `_CANONICAL_EXAMPLE_FALLBACK` string to use this declarative structure.
> 
> **As a follow-up PR, I will also consider suggesting updating `examples/bella_notte/agents/Bella_Notte_Host/instruction.txt`**, especially if we can test and validate that procedural branching performs better.
> 
> ### 2. Evidence Supporting `<subtask>` & Lowercase XML Tags
> I see that we are changing out `<state>` for `<subtask>`, along with a few other tags. I know that majority of agents uses `<subtask>`, but do we have evidence of `<subtask>` performing better than `<state>`? I have seen `<state>` work well for a customer agent.
> 
> ### 3. XML Gating is Too Aggressive (Do Not Abort Compilation)
> Currently, if the retry attempt fails schema validation, the consolidator throws an `XMLSchemaError`, which propagates through `asyncio.gather(..., return_exceptions=False)` and aborts the entire migration process. Having non-`cxas lint` approved tags does not crash the agent at runtime. The migration tool should treat XML schema validation as a **non-blocking lint warning** rather than a compilation-halting error.
> 
> * **To fix this**: If the re-prompt still returns invalid XML:
>   
>   1. Log a loud warning (e.g. `logger.warning("Synthesized XML for %s still fails schema validation after retry. Proceeding anyway.")`).
>   2. Set the status for that group in the returned dictionary to `"warning"`.
>   3. **Save the generated instructions anyway** and allow the compilation to proceed.
>   4. This avoids crashing the overall `asyncio.gather` process and ensures a single styling failure doesn't discard other successfully compiled agents.
Added changes for 1&3 in the separate commits.

For the `<subtask>` and Lowercase XML tags, subtask is a battletested approach of using and its part of our best practices. following it for consistency and until other quantifiable metric on quality.

deniscalin (approved):
Added a commit with refactoring and updates.

LGTM.

PR #229, opened 2026-05-31
feat: grouping review server

vishalghor:
##Merge after PR #227   

##
Adds a local HTTP server that the Stage 1 grouping callback can use to gate consolidation behind a user-confirmable browser UI. Wires it into MigrationService._run_stage1_consolidation as the default  callback when MigrationConfig.web_confirm_grouping is True. Default  remains False here, so no migration behavior changes unless a user
opts in via --web-confirm-grouping (or auto-confirms in headless  contexts via --auto-confirm-grouping).


deniscalin (approved):
Added commit with a few fixes and improvements.

LGTM.

PR #230, opened 2026-05-31
fix: gate file watch fallback

vishalghor:
## Merge dependencies

PR #229  needs to be merged before this

## Summary

Adds a file-watch fallback to the grouping gate: editing `<target>_grouping_plan.json` in `$EDITOR` and saving triggers the same `apply_grouping` helper as a `POST /api/grouping`. First-to-resolve wins (HTTP path and watcher converge on the same `asyncio.Event`).

Useful for `--no-web-confirm` users who prefer editing JSON in their editor, and as a safety net if the browser tab dies mid-edit.


deniscalin (approved):
Added commit with updates.

LGTM.

PR #231, opened 2026-05-31
fix: always on consolidation

vishalghor:
##Merge dependecies

merge after PR #229 , can be merged before PR #230 .

## Summary

Flips `cxas migrate` from opt-in optimization (`--optimize-for-cxas`) to always-on consolidation + browser gate. Adds `--no-consolidate` as the canonical opt-out and keeps `--no-optimize` as a one-release deprecation alias.


deniscalin (approved):
LGTM.

PR #242, opened 2026-06-02
refactor: bypass LLM for static utterances in simulation evals

ygupta89:
Currently, steps with a `static_utterance` are orchestrated by prompting the LLM to robotically repeat them. This is redundant, incurs extra tokens and latency, and can introduce instruction-following errors.

This change bypasses the LLM entirely for any active step that has a `static_utterance`. The simulator now directly marks the step as completed in the python class, sets its justification, and returns the static utterance and session/injected parameters.


AlexanderKYu (approved):
lgtm

PR #243, opened 2026-06-02
fix: catch missing graphviz dependency and exit gracefully

yuweizhan:
I got a "raise ExecutableNotFound(cmd) from e
graphviz.backend.execute.ExecutableNotFound: failed to execute PosixPath('dot'), make sure the Graphviz executables are on your systems' PATH" error when I first tried to run the migration cmd. This change allows the script to exit more gracefully while providing actionable items to users on how to resolve the error.

### DIFF
--- a/src/cxas_scrapi/migration/main_visualizer.py
+++ b/src/cxas_scrapi/migration/main_visualizer.py
@@ -47,10 +47,6 @@ from cxas_scrapi.migration.graph_visualizer import HighLevelGraphVisualizer
 from cxas_scrapi.migration.playbook_visualizer import PlaybookTreeVisualizer
 
 
-class VisualizationError(Exception):
-    """Raised when visualization fails (e.g. missing external dependency)."""
-
-
 class MainVisualizer:
     """Coordinates topology graph and detailed Rich trees with an
     interactive zoom UI (designed for Jupyter / Colab environments).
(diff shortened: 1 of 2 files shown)


vishalghor (approved):
LGTM

deniscalin (changes requested):
Thank you for sending in a fix for this!

I would request that we do this through a custom exception instead of sys.exit(1). Here's the reasoning: calling sys.exit(1) deep in a library class like MainVisualizer blocks the core program run.

If this is running in Jupyter/colab (e.g., via colab_dashboard.py) inside a notebook cell will crash the active Python kernel, abruptly destroying the user's loaded variables and active session.

I would recommend raising a custom exception (e.g. VisualizationError) or log a loud warning rather than calling sys.exit(1) inside the library class. Callers like migration_cli.py and colab_dashboard.py can then print the helpful guidance and ask if the user wants to proceed with the migration without diagrams, preventing pipeline or kernel crashes.

deniscalin (changes requested):
Thanks for making this change! I suggest one more:

Because the callers (`migration_cli.py` and `colab_dashboard.py`) do not catch this exception, it still bubbles up and crashes the execution. Also, raising it halts the method early, which prevents the Markdown report from being generated and exported, even though it doesn't need the Graphviz lib.

Instead of raising the error to halt everything, could we catch `ExecutableNotFound`, print the warning, and simply skip only the SVG export (e.g., set `svg_exported = False`), but allow the method to continue generating and exporting the Markdown report?

PR #269, opened 2026-06-08
perf(cli): implement lazy imports for CLI subcommands to improve performance

danyal-s:
### Timing Summary (on my machine)
* **Average Eager Startup Overhead (Before):** 4.88s
* **Average Lazy Startup Overhead (After):** 1.09s
* **Overall Speedup:** ~80% faster

### 🔍 Why the Overhead Was So High
In Python, importing a package compiles and executes its module-level declarations recursively. Eagerly importing the entire CLI suite dragged in 3,707 modules. On my machine, the major individual bottlenecks were:

sklearn (Scikit-Learn): 0.92 s to load, due to C-extension initializations and mathematical backend setups.
google.cloud.logging: 0.45 s, which triggers loading gRPC, OAuth credential models, and large protobuf registries.
pandas: 0.41 s for parsing dependencies, tz-aware data structures, and metadata setup.
numpy: 0.19 s.

By deferring these imports, the core CLI startup package tree dropped from 3,707 modules to just 112 modules (taking 0.07 s).

### Stats

| Command | Before (Eager) | After (Lazy) | Time Saved | Speedup |
| :--- | :--- | :--- | :--- | :--- |
| `cxas apps list` | 4.89 s | 1.09 s | 3.80 s | 348% faster |
| `cxas push` | 4.90 s | 0.66 s | 4.24 s | 643% faster |
| `cxas pull` | 4.87 s | 0.65 s | 4.22 s | 649% faster |
| `cxas lint` | 4.70 s | 0.61 s | 4.09 s | 670% faster |
| `cxas llm-lint` | 4.76 s | 2.48 s | 2.28 s | 92% faster |
| `cxas trace search` | 4.93 s | 2.99 s | 1.94 s | 65% faster |
| `cxas trace list` | 4.99 s | 2.93 s | 2.06 s | 70% faster |
| `cxas versions list` | 4.83 s | 0.69 s | 4.15 s | 600% faster |
| `cxas insights list-scorecards` | 5.04 s | 0.10 s | 4.94 s | 4940% faster |

image
image

### ✅ Verification
All unit tests (958 passed) have been verified on the perf/cli-performance branch, confirming no runtime regression.

### DIFF
--- a/src/cxas_scrapi/__init__.py
+++ b/src/cxas_scrapi/__init__.py
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, Any
 import importlib
+from typing import TYPE_CHECKING, Any
 
 # To keep the CLI startup time under 130ms, we lazy-load packages at runtime.
 # Using `if TYPE_CHECKING:` allows IDEs, type checkers, and static analyzers
 # to see the eager imports statically for full autocompletion and type-safety.
-# At runtime, TYPE_CHECKING is False, triggering the dynamic lazy loader in __getattr__.
+# At runtime, TYPE_CHECKING is False, triggering the dynamic lazy loader in
+# __getattr__.
 if TYPE_CHECKING:
     from cxas_scrapi.core.agents import Agents
     from cxas_scrapi.core.apps import Apps
@@ -106,6 +107,7 @@ else:
             module = importlib.import_module(module_path)
             return getattr(module, name)
         raise AttributeError(f"module {__name__} has no attribute {name}")
+
 
 __all__ = [
     "Agents",
--- a/src/cxas_scrapi/__init__.py
+++ b/src/cxas_scrapi/__init__.py
@@ -15,11 +15,6 @@
 import importlib
 from typing import TYPE_CHECKING, Any
 
-# To keep the CLI startup time under 130ms, we lazy-load packages at runtime.
-# Using `if TYPE_CHECKING:` allows IDEs, type checkers, and static analyzers
-# to see the eager imports statically for full autocompletion and type-safety.
-# At runtime, TYPE_CHECKING is False, triggering the dynamic lazy loader in
-# __getattr__.
 if TYPE_CHECKING:
     from cxas_scrapi.core.agents import Agents
     from cxas_scrapi.core.apps import Apps
--- a/src/cxas_scrapi/__init__.py
+++ b/src/cxas_scrapi/__init__.py
@@ -12,48 +12,138 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from cxas_scrapi.core.agents import Agents
-from cxas_scrapi.core.apps import Apps
-from cxas_scrapi.core.callbacks import Callbacks
-from cxas_scrapi.core.changelogs import Changelogs
-from cxas_scrapi.core.common import Common
-from cxas_scrapi.core.conversation_history import ConversationHistory
-from cxas_scrapi.core.deployments import Deployments
-from cxas_scrapi.core.evaluations import Evaluations
-from cxas_scrapi.core.guardrails import Guardrails
-from cxas_scrapi.core.sessions import Sessions
-from cxas_scrapi.core.tools import Tools
-from cxas_scrapi.core.variables import Variables
-from cxas_scrapi.core.versions import Versions
-from cxas_scrapi.evals.callback_evals import CallbackEvals
-from cxas_scrapi.evals.guardrail_evals import GuardrailEvals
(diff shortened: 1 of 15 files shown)


src/cxas_scrapi/core/__init__.py
```diff
@@ -14,6 +14,27 @@
 
 """Core module for CXAS Scrapi."""
 
-from .common import Common
+import importlib
+from typing import TYPE_CHECKING, Any
+
+# To keep the CLI startup time under 130ms, we lazy-load packages at runtime.
```
kmaphoenix: @danyal-s we can drop the comments here

src/cxas_scrapi/core/constants.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Common constants for the cxas_scrapi package."""
+
+DEFAULT_MODEL = "gemini-3.1-flash-live"
```
kmaphoenix: @danyal-s if this is only applicable to migration, it shouldn't belong in core.
Also, not sure that we need a file for just 1 constant

src/cxas_scrapi/__init__.py
```diff
@@ -12,48 +12,140 @@
...
-)
-from cxas_scrapi.migration.flow_visualizer import (
-    FlowDependencyResolver,
-    FlowTreeVisualizer,
-)
-from cxas_scrapi.migration.graph_visualizer import HighLevelGraphVisualizer
-from cxas_scrapi.migration.main_visualizer import MainVisualizer
-from cxas_scrapi.migration.playbook_visualizer import PlaybookTreeVisualizer
-from cxas_scrapi.utils.changelog_utils import ChangelogUtils
+# To keep the CLI startup time under 130ms, we lazy-load packages at runtime.
```
kmaphoenix: drop comments

src/cxas_scrapi/cli/insights_cli.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: PLC0415
```
kmaphoenix: @danyal-s instead of dropping these on each file, let's implement the ruff linter change at the pyproject.toml level where we can ignore the cli folder for this specific check

```
[tool.ruff.lint.per-file-ignores]
"src/cxas_scrapi/cli/**" = ["PLC0415"]
```

We can make this global exception for the CLI files to speed things up.
For the core files, we still want to maintain the standard top level imports as the standard.
However, we can make the exception on a method by method basis in the core files.
For example, in turn_evals.py w/ sklearn since it's a heavy import, we can do the inline # noqa check

kmaphoenix (changes requested):

src/cxas_scrapi/cli/insights_cli.py
danyal-s: Thanks, that's great. I'l make this change.

danyal-s (commented):
@kmaphoenix Please take a look.

Also, I found another approach to make this even cleaner, so will be putting out another PR for that to review.

EDIT: https://github.com/GoogleCloudPlatform/cxas-scrapi/pull/283

src/cxas_scrapi/cli/main.py
```diff
@@ -23,58 +25,70 @@
...
-from cxas_scrapi.core.deployments import Deployments
-from cxas_scrapi.core.evaluations import Evaluations, ExportFormat
-from cxas_scrapi.core.github import init_github_action
-from cxas_scrapi.evals.callback_evals import CallbackEvals
-from cxas_scrapi.evals.tool_evals import ToolEvals
-from cxas_scrapi.migration.config import DEFAULT_MODEL
-from cxas_scrapi.migration.dfcx_exporter import ConversationalAgentsAPI
-from cxas_scrapi.utils.eval_utils import EvalUtils
+
+DEFAULT_MODEL = "gemini-3.1-flash-live"
```
danyal-s, line 36: @kmaphoenix I have kept this here because importing it from migration config was causing heavy modules to get loaded, causing a slowdown of about 3 seconds for all commands.

src/cxas_scrapi/core/constants.py
danyal-s: done. This was added previously because main.py imported this constant which caused the whole DFCX module to get imported which caused a global slowdown. For now, I hard-coded the model into main.py to not rely on the module for a simple constant.

kmaphoenix (approved):
LGTM

PR #276, opened 2026-06-09
feat: add state-rehydration (historical_contexts) and advanced simulations guide

idlerboris:
This PR enhances the simulation evaluations by introducing the ability to pass `historical_contexts` to `simulate_conversation`, enabling state-rehydration. This is a crucial feature for jumping straight into a specific sub-agent during multi-turn simulations without needing to execute the full flow from the start.
Additionally, it adds a comprehensive guide (`docs/guides/evaluation/advanced-simulations.md`) that documents how to leverage these new state-rehydration capabilities alongside the `use_tool_fakes` feature (recently introduced upstream) for advanced, deterministic simulation testing.


docs/guides/evaluation/advanced-simulations.md
```diff
@@ -0,0 +1,50 @@
+# Advanced Simulation Testing in CXAS Scrapi
+
+With the recent additions of `historical_contexts` and `use_tool_fakes` to the Simulation Engine, you can now write robust, hermetic, and fast tests. This guide explains how these fields interact with the underlying Conversational Agents (CXAS) platform and how you can use them.
```
kmaphoenix, line 3: nit: CX Agent Studio or CXAS, to keep naming consistent through the repo

src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -419,6 +419,7 @@ def _send_request_with_retry(
         background_noise_file: str | None = None,
         burst_noise_files: list[str] | None = None,
         use_tool_fakes: bool = False,
+        historical_contexts: Optional[List[Dict[str, Any]]] = None,
```
kmaphoenix, line 422: Using Optional, List, and Dict here will raise a NameError at runtime because they are not imported in this file. Let's use modern Python 3.10 standard union generics instead (which is the standard style used elsewhere in this class):

```py
list[dict[str, Any]] | None = None
```

tests/cxas_scrapi/evals/test_simulation_evals.py
kmaphoenix, line 1: The new parameter signature needs to be accounted for in our mock assertions on mock_sessions.run, otherwise these tests fail with mismatch errors. See these tests:
- test_user_simulator
- test_user_simulator_audio
- test_simulation_evals_simulate_conversation_use_tool_fakes

tests/cxas_scrapi/evals/test_simulation_evals.py
```diff
@@ -1098,6 +1098,57 @@ def test_simulation_evals_simulate_conversation_use_tool_fakes(
...
+    app_name = "projects/test/locations/us/apps/123-abc"
+    with patch("cxas_scrapi.evals.simulation_evals.GeminiGenerate"):
+        with patch("cxas_scrapi.core.apps.AgentServiceClient"):
+            simulator = SimulationEvals(app_name=app_name)
+    test_case = {
+        "steps": [],
+        "use_tool_fakes": True,
+        "historical_contexts": [{"user": "hello"}],
+    }
+    simulator.simulate_conversation(
```
kmaphoenix, line 1129: We need to pass use_tool_fakes=True explicitly in the simulator method call, and add the noise parameter mocks to the assertions:

kmaphoenix (changes requested):

PR #280, opened 2026-06-11
feat: Add guardrail violations in simulation reports

ananyajha2000:
## Summary

- Adds guardrail violations in simulation reports. When a DFCX agent fires a guardrail during a sim run, the violation is now captured and shown inline in the HTML report at the exact conversation turn where it occurred 
- Violations appear as collapsible ❗ Guardrail Triggered dropdowns showing the type, stage, agent, and reason. No raw JSON noise.
- Run summary headers include a guardrails: N count so violations are visible at a glance without digging into the trace.
- Removes noisy "Custom Payload (output)" block from the report (retains the dropdown)

Changes

- scrapi-sim-runner.py: Added span tree traversal to extract triggered guardrails from the agent response's diagnostic info, with thread-safe per-conversation storage and guaranteed cleanup via try/finally.
- simulation_evals.py: Plumbed guardrail_details through the job result dict so the data reaches the report without a full method override.
- reporting.py: Added guardrail rendering to both the standalone and combined report paths.
- combined_report_template.html: Added the matching Jinja block for the combined report.
- base.css: Added guardrail-specific CSS classes with light and dark mode support.


elizabethhli (approved):

.agents/skills/cxas-agent-foundry/scripts/scrapi-sim-runner.py
```diff
@@ -106,7 +108,53 @@ def build_test_case(ev: dict[str, Any]) -> dict[str, Any] | None:
 
 
 class EnhancedSimRunner(SimulationEvals):
-    """Extended SimulationEvals that injects session variables."""
+    """Extended SimulationEvals that injects session variables and extracts triggered guardrails."""
+
+    def _search_guardrail_spans(self, span_dict: dict[str, Any]) -> list[dict[str, Any]]:
```
kmaphoenix, line 113: @ananyajha2000 this entire guardrail parser should be relocated into [/core/response_parser.py](https://github.com/GoogleCloudPlatform/cxas-scrapi/blob/main/src/cxas_scrapi/core/response_parser.py)

In general, we should avoid any core logic in scripts used by Skills, especially if the logic is generally reusable.

I would do the following:
- Modify `ParsedSessionResponse` class and add the new `_search_guardrail_spans` method there
- Remove `_parse_agent_response` and `_search_guardrail_spans` from the sim runner script
- Utilize the parsed payloads from the core class like:
```py
agent_text, trace_chunks, session_ended, tool_calls = self._parse_agent_response(response)
```

kmaphoenix (changes requested):

PR #283, opened 2026-06-12
perf(cli): Improve CLI performance: sub-200ms startup via Click lazy loading and typed dataclasses

danyal-s:
# Refactor CLI to Click lazy loading and strongly-typed dataclasses

## Summary

Refactored the CLI framework (`src/cxas_scrapi/cli/`) from eager `argparse` parsing to Click with lazy command loading (`LazyRootGroup`) and `@dataclass` configuration models.

This drops CLI startup and help text rendering from **~5 seconds down to ~150–190ms**, while replacing untyped `argparse.Namespace` objects across handlers with explicit, typed dataclass models (`AppPushConfig`, `RunEvalConfig`, `MigrationDfcxConfig`, etc.).

All 213 CLI unit tests pass cleanly without breaking changes to existing flags or command interfaces.

---

## Motivation

1. **Slow Startup (~5s delay)**: `main.py` previously built the entire `argparse` tree upfront. Invoking simple commands like `cxas --help` or `cxas migrate dfcx --help` forced Python to import heavy dependencies (`google.cloud.ces_v1beta`, `dialogflowcx`, `pandas`, `rich`) immediately, adding ~5s of startup delay to every command invocation.
2. **Untyped `argparse.Namespace`**: Command handlers accepted raw `argparse.Namespace` objects. Handlers couldn't easily leverage static typing/linting, making attribute typos and state mutations harder to catch early.

---

## What Changed

- **Lazy Command Loading (`main.py`)**: Replaced `argparse.ArgumentParser` with a Click `LazyRootGroup`. Commands and their dependencies are now only imported when the specific subcommand is invoked. Root `--help` is served instantly from a static summary table without importing submodules.
- **Typed Config Models (`src/cxas_scrapi/cli/*.py`)**: Defined `@dataclass` models for each command's options across all 15 CLI modules.
- **`to_dataclass` Coercion Helper (`utils.py`)**: Added `to_dataclass` to populate config dataclasses from Click context options or test mocks. Handlers accept `config: ConfigClass | Any`, preserving backward compatibility for existing unit test fixtures that pass `argparse.Namespace`.

---

## Performance Summary

Measured using `time.perf_counter()` (5 iterations per command):

image

| Command | Pre-migration (Eager `argparse`) | Post-migration (Click + Dataclasses) | Speedup |
| :--- | :---: | :---: | :---: |
| `cxas --help` | 5,014 ms | 154 ms | **32.5x faster** |
| `cxas help llm-lint` | 3,802 ms | 156 ms | **24.3x faster** |
| `cxas migrate dfcx --help` | 5,182 ms | 398 ms | **13.0x faster** |
| `cxas evals --help` | 5,014 ms | 159 ms | **31.5x faster** |
| `cxas apps --help` | 5,014 ms | 167 ms | **30.0x faster** |
| `cxas deployments --help` | 5,014 ms | 186 ms | **26.9x faster** |
| **Average Startup Latency** | **5,035 ms** | **193 ms** | **26.1x faster** |

---

## Testing & Verification

- `uv run ruff check` passes with 0 errors.
- `uv run ruff format --check .` passes (all files formatted).
- `uv run pytest tests/cxas_scrapi/cli/` — all 213 unit tests passed in 7.8s.


kmaphoenix:
@danyal-s This still seems like a good change to make.  
We've made some recent changes to `main.py` so there's some conflicts to resolve.

Can you rebase and then see if we can get this to a point where it's ready to merge?

tests/benchmarks/cli_performance_baselines.json
danyal-s, line 1: This is left intentionally here to benchmark future changes to CLI comands to ensure they do not become too slow again.

danyal-s:
@kmaphoenix I've repurposed this PR since the slow speed (which you can see in the description) was making this difficult to achieve in a reasonable time.

Using click for arg parsing should enforce some good practices across the repo and make it a lot easier to modularize + run static analyzers.

kmaphoenix:
@danyal-s thanks for digging into the CLI startup problem.
The speedup is real and we definitely want it.
A few thoughts on how to get this landed smoothly.

## 1. Rebase first
`main` has moved since this branched (evals --format json, the S008 lint rule, v1.7.0).
There are conflicts in `main.py`, `pyproject.toml`, and `uv.lock`.
Before any more changes, can you rebase onto latest `main` so we're building on current behavior?  
Or better yet, start a completely clean branch.

## 2. Narrow the Scope
A few files were changed that look unrelated to the CLI goal.
Things like `pytest.ini` (removed `testpaths`/`pythonpath`, switched to `asyncio_mode = auto` repo wide), `migration/__init__.py`, the expanded ruff ignores, and `tests/benchmarks/` widen the review surface quite a bit.
If those changes are needed, let's split them into their own PRs so each one can be reviewed on its own.

## 3. Separate lazy loading from the Click migration
I profiled startup with `python -X importtime`.
Nearly all of the cost flows through a single import: `from cxas_scrapi.utils.eval_utils import ...` at the top of `main.py`.
That one line pulls in `google.genai`, `pandas`, and `ces_v1beta` before arg parsing even starts.

As an experiment I deferred just that import on main, with argparse fully intact:

| Variant | `cxas --help` (warm) |
| :--- | :---: |
| main today | ~0.67s |
| this PR (Click) | ~0.05s |
| main + 2-line lazy import change | ~0.05s |

So we can capture the entire speedup with roughly 2-3 files changed (defer that import in `main.py`, plus your lazy `utils/__init__.py` change, which is a nice piece of this PR worth keeping).
For comparison, this PR is currently 34 files.

The Click migration itself is a much bigger conversation.
The Click tree doesn't yet cover the full argparse surface (a number of existing commands and flags aren't reachable from the new entrypoint), so merging it as is would change behavior for current users.  
I don't think we have full test coverage for all current commands and flags, so CI testing isn't catching any regressions.

<details>
<summary>Details on the coverage gap (click to expand)</summary>

The old argparse tree is still present in `main.py`, but it's unreachable.
The new `main()` calls the Click `LazyRootGroup`, and its `COMMAND_MAP` dict is now the only routing table.
Anything not in that dict returns `None` from `get_command`, so the CLI reports "No such command".
That's also why the diff is hard to review for this: no registration lines show as deleted, the commands just stop being routable.

I dumped the full command tree on both sides (every command path plus flags) and diffed them.
`main` has 87 command paths, this branch has 37.

Top-level commands not in `COMMAND_MAP` (all currently error with "No such command"):
`run`, `pull`, `push`, `export`, `push-eval`, `test-tools`, `test-callbacks`, `test-single-callback`, `ci-test`, `local-test`, `init`, `create`, `branch`, `delete`, `init-github-action`, and the `local create agent|tool|guardrail` family.

Subcommands missing from groups that did make it over:
`apps get`, `conversations get`, `versions compare`, 21 of 22 `insights` subcommands, 11 of 12 `trace` subcommands, and `callbacks`/`variables` (with `tools` moved under a new `resources` group).

The surviving commands also re-declare their flags from scratch, so most expose a subset:
- `evals report` has 5 of ~30 flags (missing `--format json` from #368, `--include`, the `--filter-*` family, `--golden-run`, `--modality`, `--output`, ...)
- `lint` has 2 of 15 (missing `--fix`, `--json`, `--only`, `--rule`, ...)
- `migrate dfcx` is missing `--target-name` and `--stage`, and renamed `--project-id` to `--gcp-project-id`, which the config layer silently drops, so `--run` mode currently can't succeed
- `trace search` has 3 of 15
- the root `--no-input` flag is gone

None of this shows up in CI because the existing tests call the handler functions directly (and `test_get_parser*` still passes against the retained argparse tree).
Only 3 tests exercise the Click entrypoint, and they only assert `--help` output.
That's the main reason I'd like a parity check before the Click swap merges.

</details>

# Proposal
Start with just the lazy loading change in a focused PR now.   
Treat the Click migration as its own effort with a parity check comparing the old and new command trees so we can prove nothing changes for users.

kmaphoenix (changes requested):
See comment re: how to break this PR up for a multi stage approach to landing these changes.

danyal-s:
I will think about doing this after we have gotten the import deferral patterns in.

PR #290, opened 2026-06-17
fix: recover from bad truncation of cuj skill.

danyal-s:
Recover from an unfortunate truncation of the CUJ skill. Should be working now.


vishalghor (approved):
LGTM

PR #312, opened 2026-06-26
test: mock Gemini centrally in offline tests

danyal-s:
Mock Gemini client: this mocks GeminiGenerate centrally in conftest.py for all offline tests (except test_gemini.py which tests the client wrapper itself).


AlexanderKYu (approved):
lgtm

PR #322, opened 2026-06-29
fix: created expectations only flag to ignore goals and updated goal …

AlexanderKYu:
…prompt to be more general


ygupta89 (changes requested):
This means that the cxas evals cli command won't get these right?

src/cxas_scrapi/prompts/llm_user_prompts.py
```diff
@@ -130,22 +130,35 @@
...
+                        step's criteria are met and it is now ready for
+                        completion."
+                *   **Case B: The Agent provided a direct answer, explanation,
+                    or statement of information (no actions required from
+                    you):** You can immediately acknowledge the response in a
+                    single turn:
+                    *   **Acknowledgment Turn:** Your `next_user_utterance`
+                        must acknowledge the information. (e.g., "Okay, I
+                        understand now. Thank you for clarifying."). The
+                        `status` REMAINS `"in progress"`. Update
```
ygupta89, line 158: Should this be in progress or completed?

ygupta89 (approved):

src/cxas_scrapi/prompts/llm_user_prompts.py
AlexanderKYu: How it is originally implemented was that the following step will parse if the justification mentioned that the goal is ready for completion.

```
*   **Case Case 3: Is the NEXT step `not started` AND the current step's
            justification includes "ready for completion"? (Transition Turn)**
            *   This is the turn to move to the next goal.
            *   Your `next_user_utterance` should introduce the next goal from
                the script.
            *   Update the *next* step's `status` to `"in progress"`.
            *   Update the *current* step's `status` to `"completed"`
```

Then it will trigger the completed status.

PR #324, opened 2026-06-30
fix(migration): avoid nested asyncio.run() in migration scripts

universe1987:
Problem: Running any migration script interactively crashed with:
```
RuntimeError: asyncio.run() cannot be called from a running event loop
```
This affected all 4 scripts (migrate.py, stage_1.py, stage_2.py, stage_3.py) because they all used asyncio.run(_run()) with async def _run(), and InquirerPy's prompt_toolkit calls its own asyncio.run() internally for interactive prompts — Python doesn't allow nesting those.

Fix: Across all 4 scripts:
  - _run() changed from async def to regular def
  - main() calls _run(args) directly instead of asyncio.run(_run(args))
  - The single async service call inside _run() gets its own asyncio.run(service.run_stage_*())

Interactive prompts now run outside any event loop, avoiding the nesting conflict.

Test:
```
python .agents/skills/cxas-dfcx-migration/scripts/migrate.py \
  --zip-file "<file_path>" \
  --project-id ces-deployment-dev \
  --location us \
  --target-name <target_name> \
  --skip-resource-selection \
  --skip-dependency-analysis \
  --yes
```

### DIFF
--- a/.agents/skills/cxas-dfcx-migration/scripts/migrate.py
+++ b/.agents/skills/cxas-dfcx-migration/scripts/migrate.py
@@ -146,7 +146,7 @@ def _build_parser() -> argparse.ArgumentParser:
 # ---------------------------------------------------------------------------
 
 
-async def _run(args) -> None:
+def _run(args) -> None:
     tracker = phase_tracker.PhaseTracker(console)
 
     # Phase 0: auth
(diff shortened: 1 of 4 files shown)


vishalghor:
- Post the original traceback so we can identify the actual offender.
- Fix the sync-that-calls-asyncio.run helper at its source (make it async and await it from _run, or expose an async variant).
- Revert _run to async def and the calls to await.
- Please check the groupingweb_review async loop and test it locally for the fix.

PR #329, opened 2026-07-01
fix(tools): add fault tolerance around retrieve_tools in Tools.get_tools_map

hongdagoogle:
Prevents initialization crashes when external OAuth/CES servers return 503/403 outages during MCP toolset discovery.

### DIFF
--- a/src/cxas_scrapi/core/tools.py
+++ b/src/cxas_scrapi/core/tools.py
@@ -191,18 +191,24 @@ class Tools(Apps):
                         )
                         tools_dict.update(openapi_tools)
                 else:
-                    toolset_tools = self.retrieve_tools(
-                        tool.name.split("/")[-1]
-                    )
-                    for toolset_tool in toolset_tools.tools:
-                        if reverse:
-                            tools_dict[toolset_tool.display_name] = (
-                                toolset_tool.name
-                            )
-                        else:
-                            tools_dict[toolset_tool.name] = (
-                                toolset_tool.display_name
-                            )
+                    try:
+                        toolset_tools = self.retrieve_tools(
+                            tool.name.split("/")[-1]
+                        )
+                        for toolset_tool in toolset_tools.tools:
+                            if reverse:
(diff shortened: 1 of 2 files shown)


kmaphoenix:
@hongdagoogle PTAL at the failed CI for linting.  

Run `uv run ruff format --check .` locally

src/cxas_scrapi/core/tools.py
```diff
@@ -191,18 +191,24 @@ def get_tools_map(self, reverse: bool = False) -> dict[str, str]:
...
+                        for toolset_tool in toolset_tools.tools:
+                            if reverse:
+                                tools_dict[toolset_tool.display_name] = (
+                                    toolset_tool.name
+                                )
+                            else:
+                                tools_dict[toolset_tool.name] = (
+                                    toolset_tool.display_name
+                                )
+                    except Exception as e:  # pylint: disable=broad-exception-caught
```
kmaphoenix: @hongdagoogle let's also make this an explicit error vs. a general catch and trying to disable the pylint here. This should suffice for both ruff checks and internal checks

```python
from google.api_core import exceptions as google_exceptions

...
(other code)
...
else:
  try:
  ...
  except google_exceptions.GoogleAPICallError as e:
```

test case will need update as well

kmaphoenix (changes requested):

hongdagoogle:
Thanks @kmaphoenix! We have formatted the codebase using ruff and pushed the updated commits (latest: f42e91a). All local checks (`ruff format --check .` and `ruff check .`) now pass cleanly.

hongdagoogle:
> @hongdagoogle PTAL at the failed CI for linting.
> 
> Run `uv run ruff format --check .` locally

Done.

src/cxas_scrapi/core/tools.py
hongdagoogle: Done.

kmaphoenix (approved):

PR #339, opened 2026-07-06
docs: Update linter rules documentation

bktsim:


ygupta89 (commented):
Do we need to update the actual lint rules or were they fixed before?

bktsim:
The linter rules already exist and work as intended, I think we just missed updating the documentation

ygupta89 (approved):

PR #358, opened 2026-07-15
test(migration): add regression tests for headless review-server boot

kmaphoenix:
## What

Adds three tests to `tests/cxas_scrapi/migration/test_service.py` covering the early grouping-review-server boot in `run_stage_1` and `run_migration`:

- `test_run_stage_1_skips_early_server_boot_in_headless_context`
- `test_run_migration_skips_early_server_boot_in_headless_context` — uses the config defaults (`web_confirm_grouping=True`) so the exact default-config path is exercised
- `test_run_migration_boots_server_early_when_interactive` — the complement, asserting the server still boots once for interactive runs

## Why

#348 added `_is_headless_context()` guards to the two early boot sites, but nothing asserts on the boot itself: the autouse `webbrowser` mock in `tests/conftest.py` would silently swallow a regression, leaving a real `ThreadingHTTPServer` binding ports during tests with no failing signal. These tests patch `boot_review_server` directly and assert it is (not) called.

## Verification

Mutation-tested: with the two `and not _is_headless_context()` guards removed from `service.py`, both skip-tests fail and the interactive test passes; with the guards intact, all pass. Full `test_service.py` module green (23 passed); ruff check/format clean.

### DIFF
--- a/tests/cxas_scrapi/migration/test_service.py
+++ b/tests/cxas_scrapi/migration/test_service.py
@@ -1139,3 +1139,178 @@ async def test_run_stage_1_auto_confirms_in_headless_context():
         )
 
     assert called["n"] == 0
+
+
+@pytest.mark.asyncio
+async def test_run_stage_1_skips_early_server_boot_in_headless_context():
+    """The early review-server boot in run_stage_1 must never run in a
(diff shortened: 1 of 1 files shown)


vishalghor (approved):
LGTM

PR #373, opened 2026-07-21
fix(migration): strip all DFCX trigger decorators and preserve original function names for Python tools

deniscalin:
This PR fixes Python tool deployment failures during Dialogflow CX migrations by resolving decorator stripping and entry function name resolution in `CodeBlockMigrator`. Previously, `visit_FunctionDef` only stripped basic decorators (`Action`, `Handler`, etc.), leaving DFCX triggers such as `BeforeagentTrigger`, `BeforeAgentTrigger`, `BeforeModelTrigger`, `PlaybookStartHandler`, and `EventTrigger` intact in the AST. When uploaded to CXAS, these unstripped decorators caused runtime `NameError` exceptions. Additionally, code blocks without explicit entry decorators fell back to hardcoding `"main"` instead of preserving the function's original name, resulting in `Function main not found` deployment errors. This update expands the decorator stripping set to cover all DFCX triggers and preserves original Python function names for entry point resolution.

### DIFF
--- a/src/cxas_scrapi/migration/code_block_migrator.py
+++ b/src/cxas_scrapi/migration/code_block_migrator.py
@@ -224,22 +224,27 @@ class ToolCallTransformer(ast.NodeTransformer):
         self.generic_visit(node)
 
         # Strip DFCX-specific decorators
+        dfcx_decorators = {
+            "Action",
+            "Handler",
+            "system",
+            "action",
+            "handler",
+            "BeforeActionTrigger",
+            "BeforeModelTrigger",
+            "BeforeagentTrigger",
+            "BeforeAgentTrigger",
+            "PlaybookStartHandler",
+            "EventTrigger",
+        }
         new_decorators = []
         for dec in node.decorator_list:
-            if isinstance(dec, ast.Name) and dec.id in [
-                "Action",
-                "Handler",
-                "system",
-                "action",
-                "handler",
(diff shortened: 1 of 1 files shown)


src/cxas_scrapi/migration/code_block_migrator.py
```diff
@@ -508,6 +513,8 @@ def _parse_code_block_with_ast(
                                 "handler",
                                 "BeforeActionTrigger",
                                 "BeforeModelTrigger",
+                                "BeforeagentTrigger",
```
kmaphoenix, line 516: typo?  Agent is lower case

kmaphoenix: actually this is duplicated 

`BeforeagentTrigger` and
`BeforeAgentTrigger`

kranthikiran01 (approved):

kmaphoenix (approved):

src/cxas_scrapi/migration/code_block_migrator.py
deniscalin: These are intended, they are the hardcoded decorators from DFCX we need to process. I have seen the typo decorators used in a recent migration from DFCX.

I'm actually authoring a slightly bigger change to commit to the same PR to get away from hardcoding. Pushing shortly.

PR #392, opened 2026-07-25
feat: add cxas-interactive-eval-reporter skill

nareshganatra:
Create an interactive report based on the input session parameters and the various error categories. The goal is to get the categories dynamically generated. 
Will need to run this on a few different agents to see how well the categorization turns out.


nareshganatra:
hey @ygupta89   @mluszczyk check this when you get a moment !  thx

.agents/skills/cxas-interactive-eval-reporter/scripts/generate_report.py
```diff
@@ -0,0 +1,824 @@
...
+                except Exception as b_err:
+                    print(f"[Warning] Batch classification error: {b_err}")
+
+        return result
+        
+    except Exception as e:
+        print(f"[Warning] LLM failure categorization skipped or offline ({e}). Using fallback clustering.")
+        return {}
+
+def fallback_categorize(expectation_text: str, justification: str) -> str:
```
mluszczyk, line 174: Do we want this fallback? It looks like a pretty simple mapping, is it useful in practice? Is this the best thing we can do when the LLM fails?

mluszczyk (commented):
left one small comment

kmaphoenix:
@nareshganatra can you clarify how this skill is differentiated from the existing `cxas-sim-eval` skill and associated reporting?

https://github.com/GoogleCloudPlatform/cxas-scrapi/tree/main/.agents/skills/cxas-sim-eval

.agents/skills/cxas-interactive-eval-reporter/scripts/generate_report.py
nareshganatra: Good catch. It's definitely basic.

Open to dropping it or simplifying further if we'd rather let them fall into a generic "Uncategorized" bucket instead. Let me know what you think!

PS - am out this week, feel free to change it if you had any suggestions, if not will look at it next week.

nareshganatra:
> @nareshganatra can you clarify how this skill is differentiated from the existing `cxas-sim-eval` skill and associated reporting?
> 
> https://github.com/GoogleCloudPlatform/cxas-scrapi/tree/main/.agents/skills/cxas-sim-eval

@kmaphoenix 
The key difference is that cxas-sim-eval focuses purely on execution and raw logging, whereas this focuses on further classification and interactive reporting.

Specifically, this adds:

Dynamic LLM Failure Clustering: Automatically groups failure modes using Gemini instead of just showing raw status.

Universal Parameter Filtering: Dynamically extracts and builds filter controls for all session parameters so you can slice and dice the data.

Essentially, cxas-sim-eval runs the tests, and this turns those results into an interactive, categorized diagnostic dashboard.

There are a couple of reports that I had shared in our group chat and can share the same again next week (out this week) or if @mluszczyk  get a chance he might be able to share it with you.

nareshganatra:
Closing this draft PR as this capability has been consolidated into the `cxas-agent-foundry` skill and merged upstream in #429.

PR #399, opened 2026-07-29
fix(path_resolution): normalize path separators for file references on windows

corbincalvert:
The `_resolve_paths` helper now normalizes `base_path` separators to forward slashes. This ensures that `extra_prefixes` matching works correctly on Windows systems, preventing `FileNotFoundError` when referencing files.


kmaphoenix:
Hi @corbincalvert, thanks for the submission!

Couple of things:
- Please run the ruff linter / formatters locally to fix the lint errors
- Please add unit tests to verify that both Windows and non-windows paths are properly parsed

src/cxas_scrapi/utils/lint_rules/schema.py
```diff
@@ -1,391 +1,396 @@
...
+        if base_path:
+            alt = Path(base_path) / data
+            if alt.exists():
+                path_to_check = alt
+                resolved = True
+
+        if not resolved and extra_prefixes:
+            # Normalize separators so prefix matching works on Windows,
+            # where base_path contains backslashes but references use "/".
+            base_norm = (
```
kmaphoenix: Let's add unit tests to cover both Windows and non-Windows cases here

kmaphoenix (changes requested):

kmaphoenix (changes requested):
@corbincalvert it also looks like the changes to the file were submitted using windows CRLF line endings.  This causes the entire file to appear as changed in the history, when only a few lines were actually changed.

Please run the following command locally to revert the CRLF (`\r\n`) line endings to standard Unix LF ('\n') line endings so we can avoid polluting `git blame` 

`dos2unix src/cxas_scrapi/utils/lint_rules/schema.py`

Alternatively you should be able to set your editor / IDE to use LF line endings

corbincalvert:
Thanks for the review! Added unit tests in tests/cxas_scrapi/utils/test_lint_rules.py covering _resolve_paths path parsing:

POSIX base path — a forward-slash base_path resolves a referenced instruction file via the extra_prefixes fallback.
Windows base path — the same reference resolves when base_path contains backslashes (e.g. ...\agents\My_Agent). This is the regression case: it fails with FileNotFoundError against the pre-fix logic, since prefix matching compared the forward-slash reference against the backslash path.
Nested structures — references inside dict payloads resolve with a Windows base path (mirrors real agent config shape).
Missing file — a nonexistent reference still raises FileNotFoundError, confirming the fix doesn't loosen validation.

All four pass with pytest tests/cxas_scrapi/utils/test_lint_rules.py -k resolve_paths.

corbincalvert:
> @corbincalvert it also looks like the changes to the file were submitted using windows CRLF line endings. This causes the entire file to appear as changed in the history, when only a few lines were actually changed.
> 
> Please run the following command locally to revert the CRLF (`\r\n`) line endings to standard Unix LF ('\n') line endings so we can avoid polluting `git blame`
> 
> `dos2unix src/cxas_scrapi/utils/lint_rules/schema.py`
> 
> Alternatively you should be able to set your editor / IDE to use LF line endings

Fixed this.

corbincalvert:
Is there an action item waiting on me? Please review.

PR #400, opened 2026-07-29
feat(evals): make vertex_location configurable in SimulationEvals and TurnEvals

chees:
## Summary
Currently, `SimulationEvals` and `TurnEvals` hardcode `vertex_location = "global"` when initializing `GeminiGenerate`. Organizations with strict regional data residency policies (such as requiring `europe-west4` or `us-central1` for Vertex AI calls) are unable to use generative evaluation capabilities without manually modifying installed library code.

This PR makes `vertex_location` fully configurable:
1. **Explicit Constructor Parameter**: `SimulationEvals` and `TurnEvals` now accept an optional `vertex_location: str | None = None` parameter.
2. **Environment Variable Fallback**: If `vertex_location` is not explicitly provided, it falls back to the `VERTEX_LOCATION` environment variable.
3. **Backwards Compatibility**: Defaults to `"global"` if neither explicit argument nor environment variable is provided, preserving existing behavior.

## Key Changes
- **`src/cxas_scrapi/evals/simulation_evals.py`**: Added `vertex_location` argument to `SimulationEvals.__init__` and resolve via `vertex_location or os.getenv("VERTEX_LOCATION", "global")`.
- **`src/cxas_scrapi/evals/turn_evals.py`**: Added `vertex_location` argument to `TurnEvals.__init__` and resolve via `vertex_location or os.getenv("VERTEX_LOCATION", "global")`.
- **`tests/cxas_scrapi/evals/test_simulation_evals.py`**: Added unit test cases for constructor parameter and environment variable resolution.
- **`tests/cxas_scrapi/evals/test_turn_evals.py`**: Added unit test cases for constructor parameter and environment variable resolution.

## Verification
Ran test suite:
```bash
uv run pytest tests/cxas_scrapi/evals/test_simulation_evals.py tests/cxas_scrapi/evals/test_turn_evals.py
```
Result: 67 passed with 0 errors.

### DIFF
--- a/src/cxas_scrapi/cli/migration_cli.py
+++ b/src/cxas_scrapi/cli/migration_cli.py
@@ -626,7 +626,9 @@ class MigrationCLI:
 
             migration_service = MigrationService(
                 project_id=config.project_id,
-                location="us",
+                location=config.location,
+                gemini_location=config.vertex_location
+                or os.getenv("VERTEX_LOCATION", "global"),
                 default_model=config.model,
             )
 
--- a/src/cxas_scrapi/cli/migration_cli.py
+++ b/src/cxas_scrapi/cli/migration_cli.py
@@ -627,8 +627,7 @@ class MigrationCLI:
             migration_service = MigrationService(
                 project_id=config.project_id,
                 location=config.location,
-                gemini_location=config.vertex_location
-                or os.getenv("VERTEX_LOCATION", "global"),
+                gemini_location=config.vertex_location,
                 default_model=config.model,
             )
 
--- a/src/cxas_scrapi/evals/simulation_evals.py
+++ b/src/cxas_scrapi/evals/simulation_evals.py
@@ -19,7 +19,6 @@ import enum
 import functools
 import inspect
 import json
-import os
 import re
 import shutil
 import time
@@ -433,15 +432,12 @@ class SimulationEvals(Apps):
 
         # Vertex AI requires a specific region (e.g. global), whereas CXAS
         # Apps use 'us' or 'eu'
-        self.vertex_location = vertex_location or os.getenv(
-            "VERTEX_LOCATION", "global"
-        )
-
         self.genai_client = GeminiGenerate(
             project_id=self.project_id,
-            location=self.vertex_location,
+            location=vertex_location,
             credentials=self.creds,
         )
+        self.vertex_location = self.genai_client.location
 
     def _parse_agent_response(
         self, response: Any
--- a/src/cxas_scrapi/evals/simulation_evals.py
+++ b/src/cxas_scrapi/evals/simulation_evals.py
@@ -415,7 +415,7 @@ class SimulationEvals(Apps):
             expectations_only: Whether to run expectations only
             deployment_id: Optional deployment ID
             vertex_location: Optional Vertex AI location (defaults to
-              VERTEX_LOCATION env var or 'global')
+                VERTEX_LOCATION env var or 'global')
         """
         self.app_name = app_name
         self.expectations_only = expectations_only
--- a/src/cxas_scrapi/evals/simulation_evals.py
+++ b/src/cxas_scrapi/evals/simulation_evals.py
@@ -404,7 +404,7 @@ class SimulationEvals(Apps):
         rate_limiter: RateLimiter | None = None,
         expectations_only: bool = False,
         deployment_id: str | None = None,
-        vertex_location: str | None = None,
+        vertex_location: str = "global",
         **kwargs: typing.Any,
     ) -> None:
         """Initializes the SimulationEvals client.
@@ -414,8 +414,7 @@ class SimulationEvals(Apps):
             rate_limiter: Optional RateLimiter for API calls
             expectations_only: Whether to run expectations only
             deployment_id: Optional deployment ID
-            vertex_location: Optional Vertex AI location (defaults to
-                VERTEX_LOCATION env var or 'global')
+            vertex_location: Optional Vertex AI location. Defaults to 'global'.
         """
         self.app_name = app_name
         self.expectations_only = expectations_only
@@ -437,7 +436,6 @@ class SimulationEvals(Apps):
             location=vertex_location,
             credentials=self.creds,
         )
-        self.vertex_location = self.genai_client.location
 
     def _parse_agent_response(
         self, response: Any
--- a/src/cxas_scrapi/evals/simulation_evals.py
+++ b/src/cxas_scrapi/evals/simulation_evals.py
@@ -408,6 +408,16 @@ class SimulationEvals(Apps):
         vertex_location: str | None = None,
         **kwargs: typing.Any,
     ) -> None:
+        """Initializes the SimulationEvals client.
+
+        Args:
+            app_name: CXAS App Name
+            rate_limiter: Optional RateLimiter for API calls
+            expectations_only: Whether to run expectations only
+            deployment_id: Optional deployment ID
+            vertex_location: Optional Vertex AI location (defaults to
+              VERTEX_LOCATION env var or 'global')
+        """
         self.app_name = app_name
         self.expectations_only = expectations_only
         project_id = app_name.split("/")[1]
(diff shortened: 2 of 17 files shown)


kmaphoenix:
@chees please run `uv run ruff check .` locally to fix the lint errors.

src/cxas_scrapi/cli/migration_cli.py
```diff
@@ -626,7 +626,9 @@ def run(self, default_agent_name: str, cx_api: Any) -> None:
 
             migration_service = MigrationService(
```
kmaphoenix, line 627: TAL at `MigrationService.__init__` in `src/cxas_scrapi/migration/service.py` where `gemini_location` is initialized. Current set to `global` which could negate the `VERTEX_LOCATION` arg from the os.

src/cxas_scrapi/utils/changelog_utils.py
```diff
@@ -344,7 +345,7 @@ def summarize_changelogs(
             else:
                 cl = GeminiGenerate(
                     project_id=project_id,
-                    location="us-central1",
+                    location=os.getenv("VERTEX_LOCATION", "us-central1"),
```
kmaphoenix: Should add an optional `vertex_location` to this method. Otherwise, the only way to set this location here is via env vars

kmaphoenix (changes requested):
In addition to the inline changes, you should consolidate all of your changes where `GeminiGenerate` is being created and instead move the code check directly into `src/cxas_scrapi/utils/gemini.py`

This will prevent the duplication of the same code over and over:
```py
class GeminiGenerate:
    def __init__(
        self,
        project_id: str,
        location: str | None = None,
        credentials: typing.Any = None,
        model_name: str = "gemini-3.1-pro-preview",
        max_concurrent_requests: int = 3,
    ) -> None:
        self.location = location or os.getenv("VERTEX_LOCATION", "global")
```

src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -404,8 +404,19 @@ def __init__(
         rate_limiter: RateLimiter | None = None,
         expectations_only: bool = False,
         deployment_id: str | None = None,
+        vertex_location: str | None = None,
```
kmaphoenix: This should default to `global` for backwards compat

src/cxas_scrapi/evals/simulation_evals.py
```diff
@@ -421,13 +432,12 @@ def __init__(
...
         # Vertex AI requires a specific region (e.g. global), whereas CXAS
         # Apps use 'us' or 'eu'
-        vertex_location = "global"
-
         self.genai_client = GeminiGenerate(
             project_id=self.project_id,
             location=vertex_location,
             credentials=self.creds,
         )
+        self.vertex_location = self.genai_client.location
```
kmaphoenix: This seems unnecessary. It's not used anywhere else in this file.

src/cxas_scrapi/evals/turn_evals.py
```diff
@@ -130,13 +130,16 @@ def __init__(
         app_name: str,
         creds: typing.Any = None,
         rate_limiter: RateLimiter | None = None,
+        vertex_location: str | None = None,
```
kmaphoenix: Should default to `global` for backwards compat

src/cxas_scrapi/evals/turn_evals.py
```diff
@@ -149,13 +152,12 @@ def __init__(
...
         # Initialize GenAI Client
         project_id = app_name.split("/")[1]
-        vertex_location = "global"
-
         self.genai_client = GeminiGenerate(
             project_id=project_id,
             location=vertex_location,
             credentials=self.creds,
         )
+        self.vertex_location = self.genai_client.location
```
kmaphoenix: remove

src/cxas_scrapi/migration/service.py
```diff
@@ -129,7 +129,7 @@ def __init__(
         self,
         project_id: str,
         location: str = "global",
-        gemini_location: str = "global",
+        gemini_location: str | None = None,
```
kmaphoenix: should default to `global`

src/cxas_scrapi/migration/service.py
```diff
@@ -167,6 +167,7 @@ def __init__(
             model_name="gemini-3.1-pro-preview",
             max_concurrent_requests=15,
         )
+        self.gemini_location = self.gemini_client.location
```
kmaphoenix: remove

src/cxas_scrapi/utils/tracing/trace_config.py
```diff
@@ -63,6 +63,8 @@ class GeminiConfig(BaseModel):
     # `gemini-3-flash-preview` supports both text and audio Parts and the
     # ThinkingConfig API. Override per-project via `trace.yaml: gemini.model`.
     model: str = "gemini-3-flash-preview"
+    # Vertex AI location for Gemini calls (e.g. "europe-west4", "global").
+    location: str | None = None
```
kmaphoenix: default to `global`

src/cxas_scrapi/utils/changelog_utils.py
```diff
@@ -255,6 +256,7 @@ def summarize_changelogs(
         vertex_client_or_project: Any,
         changelogs: list[dict[str, Any]],
         project_id: str | None = None,
+        vertex_location: str | None = None,
```
kmaphoenix: default to `us-central1` for backwards compat

src/cxas_scrapi/utils/gemini.py
```diff
@@ -30,7 +31,7 @@ class GeminiGenerate:
     def __init__(
         self,
         project_id: str,
-        location: str = "global",
+        location: str | None = None,
```
kmaphoenix: default to `global` for backwards compat

src/cxas_scrapi/utils/changelog_utils.py
```diff
@@ -332,7 +334,15 @@ def summarize_changelogs(
         try:
             # Handle if the user passes the vertex framework client or strings
 
-            if isinstance(vertex_client_or_project, GeminiGenerate):
+            is_gemini_gen = False
```
kmaphoenix: This entire block needs to be rewritten.
Currently it appears it's written to simply pass the unit test.

A safer approach for the code and unit tests is to check `hasattr` on the incoming variable `vertex_client_or_project`

Sample proposal
```py
            if hasattr(vertex_client_or_project, "generate"):
                response_text = vertex_client_or_project.generate(
                    prompt=prompt, model_name="gemini-2.5-flash"
                )
            elif hasattr(vertex_client_or_project, "models"):
                response = vertex_client_or_project.models.generate_content(
                    model="gemini-2.5-flash", contents=prompt
                )
                response_text = response.text
            else:
                loc = vertex_location or os.getenv(
                    "VERTEX_LOCATION", "us-central1"
                )
                cl = GeminiGenerate(
                    project_id=project_id,
                    location=loc,
                    model_name="gemini-2.5-flash",
                )
                response_text = cl.generate(prompt=prompt)
```

kmaphoenix (changes requested):
@chees in addition to the inline comments / change, make sure that you are always running both the ruff linter and the ruff formatter locally.

- Lint ->  `uv run ruff check .`
- Format -> `uv run ruff format --check .`

src/cxas_scrapi/evals/simulation_evals.py
chees: It is, cause the default is defined in GeminiGenerate

src/cxas_scrapi/evals/turn_evals.py
chees: It is, cause the default is defined in GeminiGenerate

src/cxas_scrapi/utils/gemini.py
chees: that gets done on line 57:

`self.location = location or os.getenv("VERTEX_LOCATION", "global")`

PR #462, opened 2026-09-17
feat(evals): add optional Naturalness Metric to SimulationEvals

kranthikiran01:
## What

Adds an **optional** Naturalness Metric to `SimulationEvals`. It grades how
human-like an agent sounds over the course of a simulation — useful for voice
agents, where a functionally correct turn can still sound robotic.

Existing behaviour is untouched. A test case with no `naturalness_metric` key
makes no extra model call and gains no extra result keys.

## How it grades

Every agent turn gets a label (`Bot-like` | `Transitional` | `Human-like`), a
1–5 score, and a list of structured factors:

```json
[
  {"quality": "emotion", "score": 4, "value": "warm",
   "reason": "Opened with [warm] and slowed delivery once the caller sounded frustrated."},
  {"quality": "spokenNumbers", "score": 1, "value": "digits",
   "reason": "Read '150' as digits instead of 'a hundred and fifty pounds'."}
]
```

**Default turn qualities:** `emotion`, `pacing`, `grammarStyle`, `disfluency`,
`lexicalVariety`, `concision`, `spokenNumbers`, `empathyCalibration`,
`scriptedness`, `turnTaking`.

**Default conversation qualities:** `personaConsistency`,
`repetitionAvoidance`, `emotionalArcTracking`, `conversationalFlow`.

The rubric rewards natural human speech — contractions, fragments,
self-corrections, fillers, varied sentence length, emotive and pacing tags,
numbers spoken out — and penalises textbook-perfect, scripted prose.

## Scoring

A turn's score is the **mean of its own factor scores**, so the score and label
can never contradict the cited evidence.

```
overall = turn_weight * mean(turn scores)
        + (1 - turn_weight) * mean(conversation factor scores)
```

It collapses to the turn mean when there are no conversation factors. `5` means
fully human-like. Labels come from the configurable bands `bot_like_below`
(default `2.5`) and `human_like_at_or_above` (default `3.75`).

## Configuration

Declared per test case under `naturalness_metric` (aliases: `naturalness`,
`naturalness_config`).

```yaml
# Shorthand — enable with defaults
- name: upset_caller_missed_delivery
  naturalness_metric: true

# Full form
- name: upset_caller_missed_delivery
  naturalness_metric:
    turn_weight: 0.7
    human_like_at_or_above: 4.0
    pass_threshold: 3.5
    extra_guidance: |
      This agent should slow down and soften once the caller escalates.
```

- `common_naturalness_metric` sets file-wide defaults, merged key-by-key under
  each eval's own block.
- A whole run can be graded without editing YAML via the `--naturalness` CLI
  flag, or the `naturalness` kwarg on `SimulationEvals`, `run_simulations`,
  `simulate_conversation` and `run_all_evals`.
- `pass_threshold` is unset by default, which makes the metric purely
  informational. When set, a run fails if the overall score falls below it.

## Compatibility

- Fully backward and forward compatible — no `naturalness_metric` key means no
  model call and no new result keys.
- Unknown config keys are tolerated; malformed config is treated as disabled
  rather than raising.
- Every grading failure path (empty trace, empty model output, an exception) is
  logged and returns `None`, so grading can never break a simulation run.

## Reporting

When the metric ran, each sim result gains `naturalness`, `naturalness_label`
and `naturalness_details`. These keys are absent otherwise. The combined HTML
report shows a naturalness badge on the run summary plus a collapsible per-turn
breakdown with each factor's justification.

## Cost

One extra Gemini call per simulation run, only when the metric is enabled.

## Testing

- 28 unit tests for the new module: turn extraction, config parsing
  (back/forward compat, shorthand, malformed input), factor-mean scoring,
  blending, thresholds, prompt contents, JSON serialisability, and all three
  failure modes.
- Integration tests for `SimulationEvals` and for the reporting merge and HTML
  render paths.
- `ruff check`, `ruff format --check` clean; full suite green.
- Also performed a live test against a deployed agent and verified the emitted
  scores, labels and factor justifications.


.agents/skills/cxas-agent-foundry/assets/project-template/evals/simulations/example_simulations.yaml
```diff
@@ -118,3 +120,77 @@ evals:
...
+  #
+  # Shorthand: `naturalness_metric: true` enables it with all defaults.
+
+  - name: upset_caller_billing_dispute
+    tags: [P0, voice, naturalness, CUJ-3]
+    naturalness_metric:
+      # false keeps the block in the file but skips grading. Default: true
+      enabled: true
+      # Grading model. Omit to reuse the simulation's eval_model
+      model: "gemini-2.5-pro"
```
ygupta89: lets use a more current model as an example

ygupta89 (approved):

.agents/skills/cxas-agent-foundry/assets/project-template/evals/simulations/example_simulations.yaml
kranthikiran01: Good catch! Updated to more recent models

